Issue #964, opened 2020-11-27
ScanMode should not offer the mode "listen" anymore as it is automatically created


jfhenon:
![image](https://user-images.githubusercontent.com/20402845/100479065-6aac6f00-30ed-11eb-8600-8eeebb7ca9f2.png)

PR #971, opened 2020-12-02
Remove listen#964

attilasuto:
ScanMode should not offer the mode "listen" anymore as it is automatically created #964

### DIFF
--- a/src/client/Engine/ScanModes.spec.jsx
+++ b/src/client/Engine/ScanModes.spec.jsx
@@ -65,11 +65,11 @@ describe('ScanModes', () => {
         scanModes={testConfig.engine.scanModes}
       />, container)
     })
-    Simulate.change(document.getElementById('engine.scanModes.0.scanMode'), { target: { value: 'listen' } })
+    Simulate.change(document.getElementById('engine.scanModes.0.scanMode'), { target: { value: 'every1Min' } })
     expect(dispatchNewConfig).toBeCalledWith({
       type: 'update',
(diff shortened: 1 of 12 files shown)


jfhenon (changes requested):
This needs a "migration" of the json on the backend.

burgerni10:
Before managing the migration, please rebase (and not merge) your branch to be up to date with the repo.

jfhenon:
need branch rebase and then write a migration to remove "listen" scanMode in it exists in oibus.json

burgerni10:
@attilasuto I made the rebase. Your branch is now up to date with master, you can work on the migration :)

jfhenon:
> @attilasuto I made the rebase. Your branch is now up to date with master, you can work on the migration :)

thankqs @burgerni10. The migration is done on the backend by @kukukk

kukukk:
After some further discussion we need to:
1. keep the scanModes setting as it is now. We don't need listen to be defined.
2. keep the South point config as it is when listen is not supported (Modbus) - because it is not in the scanModes list, it won't apear
3. hard-code for MQTT the listen scanMode instead of showing the real scanModes list
4. we can keep the check in the Engine, but it won't do anything because if the listen scanMode is not in the scanModes list, it will be ignored.
5. the migration needs only to remove the listen scanMode from the scanModes list, but do not touch the points section.

Step 3) still needs to be implemented on frontend and step 5) on backend.

jfhenon (approved):

PR #973, opened 2020-12-04
Issue #972: Initialize Cache in safe mode

### LINKED ISSUE #972: Bug when starting a fresh installation
**Describe the bug**
Error 500 when starting the admin interface

**To Reproduce**
1. Start a fresh installation (so it will start in safe mode)
2 Launch the admin interface: 
**Screenshots**
image

**Desktop (please complete the following information):**

kukukk:
Fix #972

### DIFF
--- a/src/engine/Cache.class.js
+++ b/src/engine/Cache.class.js
@@ -87,7 +87,17 @@ class Cache {
    * @param {object} activeApis - The active North applications
    * @return {void}
    */
-  async initialize(activeApis) {
+  async initializeApis(activeApis) {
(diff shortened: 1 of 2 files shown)


jfhenon:
Do we need to initialize the cache or just not use it in safe mode?

kukukk:
We show some Cache related information on the Engine's health page which is available also in safe mode.

jfhenon:
This is the point. Do we need the cache information when we are in safe mode? not sure but I guess it does not hurt (hopefully, the access to the server will not break because of the cache).

jfhenon (approved):

PR #994, opened 2020-12-18
WIP: North CsvToHtpp

victor-tndl:


victor-tndl:
@jfhenon @burgerni10 I pushed my first version of the CsvToHttp North. Let me know what do you think about it and what kind of tests do you expect.

I may start the test this afternoon if you give me a response. However, I'll start Monday.

burgerni10 (approved):

burgerni10 (approved):

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,241 @@
...
+            return ApiHandler.STATUS.LOGIC_ERROR
+          },
+          step: (step) => {
+            csvDataParsed.push(step.data)
+          },
+          complete: () => {
+            // The csv parsing is a success so we begin to map it
+            // Initialize the body to an empty array
+            this.httpBody = []
+            this.convertToHttpBody(csvDataParsed)
```
burgerni10, line 85: resolve should be juste after this function

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,241 @@
...
+          },
+        })
+        resolve()
```
burgerni10, line 88: I don't think the resolve of the promise should be used here, but rather once the complete callback of papaparser is called

src/north/CsvToHttp/CsvToHttp.class.js
burgerni10, line 85: convertToHttpBody is async (it is a promise) so this call should be used with await

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,241 @@
...
+        if (jsonObject.length > 0) {
+          jsonObject.forEach((jsonObj) => {
+            const response = this.convertCSVRowIntoHttpBody(jsonObj)
+            // Test the result of the mapping/convertion before add it in the httpBody
+            if (response) {
+              this.httpBody.push(response)
+            }
+          })
+        }
+        this.logger.silly('End of mapping (csv file to http body)')
```
burgerni10, line 112: this log does not seem useful to me : in case of success, the final log is used, if not an error log is used

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,241 @@
...
+
+    const { applicativeHostUrl, requestMethod, authentication, csvDelimiter, mapping, proxy = null } = applicationParameters.CsvToHttp
+
+    // Creation of an entity request with is composed of all necessaries informations
+    this.request = {}
+    this.request.host = applicativeHostUrl
+    this.request.method = requestMethod
+    this.request.authentication = authentication
+
+    this.mapping = mapping
```
burgerni10, line 28: mapping does not seem to exist in a new config. You can use this.mapping = mapping || {}

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  /**
+   * @param {Object} jsonObject - jsonObj
+   * @return {Object} - The send status
+   */
+  convertCSVRowIntoHttpBody(jsonObject) {
+    const object = {}
+    this.mapping.forEach((mapping) => {
+      if (Object.prototype.hasOwnProperty.call(jsonObject, mapping.csvField)) {
```
burgerni10, line 128: This test could be clearer with : if (jsonObject[mapping.csvField])

src/north/CsvToHttp/CsvToHttp.schema.jsx
```diff
@@ -0,0 +1,107 @@
...
+  },
+  authentication: { type: 'OIbAuthentication' },
+  CsvToHttpParameters: {
+    type: 'OIbTitle',
+    label: 'Mapping Parameters',
+    children: (
+      <div>
+        <ul>
+          <li>
+            The unselected index will be obmitted during the mapping
```
burgerni10, line 46: I think that "index" here is confusing : maybe you mean tag or key or column name ? For me index is the row number

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,241 @@
...
+   */
+  convertToCorrectType(value, type) {
+    let valueConverted
+    switch (type) {
+      case 'integer':
+        try {
+          valueConverted = parseInt(value)
+        } catch (e) {
+          this.logger.error(e)
+          valueConverted = value
```
burgerni10, line 152: I don't know what your use case is, but is it ok to keep a bad typed value ?

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,241 @@
...
+
+      // TODO: Improve the construction of the requests (this.request)
+      // This part could be simplify thanks to interaction between react components
+      if (this.request.authentication.type === 'Basic') {
+        this.logger.silly('Sending basic request')
+
+        await this.engine.sendRequest(this.request.host, this.request.method, this.request.authentication, this.proxy,
+          JSON.stringify(body), baseHeaders)
+      }
+      if (this.request.authentication.type === 'Custom') {
```
burgerni10, line 229: should be else if

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,241 @@
...
+   */
+  async sendRequest(body) {
+    try {
+      // Create a base header in order not to send a request with a file but a content
+      const baseHeaders = { 'Content-Type': 'application/json' }
+
+      // TODO: Improve the construction of the requests (this.request)
+      // This part could be simplify thanks to interaction between react components
+      if (this.request.authentication.type === 'Basic') {
+        this.logger.silly('Sending basic request')
```
burgerni10, line 224: Sending request with Basic Auth

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,241 @@
...
+
+        await this.engine.sendRequest(this.request.host, this.request.method, this.request.authentication, this.proxy,
+          JSON.stringify(body), baseHeaders)
+      }
+      if (this.request.authentication.type === 'Custom') {
+        this.logger.silly('Sending request with heaaders')
```
burgerni10, line 230: Sending request with Custom Auth Headers

burgerni10 (changes requested):

burgerni10:
I think I get the overall logic. These are minor changes. Please contact me if you need more details !

src/north/CsvToHttp/CsvToHttp.class.js
victor-tndl: What do you mean by bad typed value?

burgerni10: If this code is executed, it means the value was not an Integer as expected. In this case, is it correct to keep the value (which is not an integer) nevertheless, event if you are supposed to receive an integer on the server side ?

src/services/request.service.js
```diff
@@ -220,6 +220,10 @@ const sendRequest = async (engine, requestUrl, method, authentication, proxy, da
     const basic = Buffer.from(`${authentication.username}:${decryptedPassword}`).toString('base64')
     headers.Authorization = `Basic ${basic}`
   }
+  if (authentication && (authentication.type === 'Custom')) {
```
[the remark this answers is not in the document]
victor-tndl: With eslint the structure ```
if (...) {
 ...
 } else {
    if (...) {
      ...
    }
 }
``` is not allowed. Moreover if others authentications are added the structure ```
if (...) {
 ...
 }
if (...) {
 ...
 }
```  seems better for me.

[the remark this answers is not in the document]
Comment (burgerni10) [reply src/services/request.service.js:223]: Note `else { if () { ... } }` but just `else if () { ... }`

In this specific case, the best way could be 
```
if (authentication) {
   if (authentication.type ==='Basic') { ... }
   else if (authentication.type === 'Custom') {...}
}
```
I think this way of writing these conditions is the best for now : it will be cleaner to go for a switch when we'll have more authentication type to manage.

File: src/north/CsvToHttp/CsvToHttp.class.js (hunk 7 above)
Comment (victor-tndl) [reply src/north/CsvToHttp/CsvToHttp.class.js:152]: Not sure to understand your point of view, but for me we send a Json in the body, so type could be important but some value as int can be either string or integer in Json. Moreover, if a user wants only string and none typed value as int, float... it will be possible for him.

File: src/north/CsvToHttp/CsvToHttp.class.js (hunk 2 above)
Comment (victor-tndl) [reply src/north/CsvToHttp/CsvToHttp.class.js:88]: `convertToHttpBody` is not anymore a async function (it was one with csvtojson parser)

File: src/north/CsvToHttp/CsvToHttp.class.js (hunk 1 above)
Comment (victor-tndl) [reply src/north/CsvToHttp/CsvToHttp.class.js:85]: `convertToHttpBody` is not anymore a async function (it was one with csvtojson parser)

File: src/north/CsvToHttp/CsvToHttp.class.js (hunk 2 above)
Comment (victor-tndl) [reply src/north/CsvToHttp/CsvToHttp.class.js:88]: Do you prefer that `parseAndMapp` don't return a Promise?

File: src/north/CsvToHttp/CsvToHttp.class.js (hunk 7 above)
Comment (burgerni10) [reply src/north/CsvToHttp/CsvToHttp.class.js:152]: Yes indeed, for Json spec, it is not a problem. My point if more for your application. I was wondering how your endpoint would manage a bad data. 
In this example, if instead of an integer there is a value like 'NA' or just a random string, it will be kept in the field that will be sent to your endpoint anyway.
In other words, I am not sure to see the point of this function if when you have a bad data in input, you keep the data anyway ?

File: src/north/CsvToHttp/CsvToHttp.class.js (hunk 1 above)
Comment (burgerni10) [reply src/north/CsvToHttp/CsvToHttp.class.js:85]: Ok, I was confused because of the use of a promise : maybe you can remove the promise from convertToHttpBody ?

Comment (victor-tndl) [reply src/north/CsvToHttp/CsvToHttp.class.js]: I removed it

File: src/north/CsvToHttp/CsvToHttp.class.js (hunk 2 above)
Comment (burgerni10) [reply src/north/CsvToHttp/CsvToHttp.class.js]: Papa.parse seems async to me : https://www.papaparse.com/docs#local-files

If you remove the promise or keep the resolve where it is now, what will happen is that the resolve() can fire before the csv is completely parse, so even before the complete() of Papa.parse is called. In this case you won't have any data to send

File: src/north/CsvToHttp/CsvToHttp.class.js (hunk 7 above)
Comment (victor-tndl) [reply src/north/CsvToHttp/CsvToHttp.class.js:144]: The aim is to convert data to one chosen type data. For example to convert a string into a date or a string into a integer/float. This examples would be the more commons.

Comment (burgerni10) [reply src/north/CsvToHttp/CsvToHttp.class.js:144]: I see. Can you explicit this behaviour on the documentation please ? So the user will know that the data are kept in the rare case of a bad typed value.

File: src/services/request.service.js (hunk 11 above)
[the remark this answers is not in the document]
Comment (jfhenon) [reply src/services/request.service.js]: My view is that if we start to have several types: 

````
switch (type) {
    case 'basic': .... 
````
is more readable to me.
JF

File: src/north/CsvToHttp/CsvToHttp.class.js (hunk 4 above)
Comment (jfhenon) [reply src/north/CsvToHttp/CsvToHttp.class.js]: or mapping ?? {}

File: src/north/CsvToHttp/CsvToHttp.class.js [hunk 12]
```diff
@@ -0,0 +1,230 @@
...
+   */
+  async sendData() {
+    return new Promise((resolve, reject) => {
+      try {
+        const maxLength = 100
+
+        if (this.httpBody.length > maxLength) {
+          // Divide the current body in array of maximun maxLength elements
+          let i = 0
+          while (i < this.httpBody.length) {
```
jfhenon, line 188: OK, this one is not your fault but my phobia for "while" loops :-) 
look at https://www.iditect.com/how-to/50650976.html to find inspiration

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,230 @@
...
+   * @param {Array} body - Body to send
+   * @return {void}
+   */
+  async sendRequest(body) {
+    try {
+      // Create a base header in order not to send a request with a file but a content
+      const baseHeaders = { 'Content-Type': 'application/json' }
+
+      if (this.request.authentication.type === 'Basic') {
+        this.logger.silly('Sending request with Basic Authentication')
```
jfhenon, line 214: should also be a switch but in both case, it seems you use the same 
`await this.engine.sendRequest(this.request.host, this.request.method, this.request.authentication, this.proxy, JSON.stringify(body),
          baseHeaders)`

src/north/CsvToHttp/CsvToHttp.schema.jsx
```diff
@@ -0,0 +1,107 @@
...
+  },
+  authentication: { type: 'OIbAuthentication' },
+  CsvToHttpParameters: {
+    type: 'OIbTitle',
+    label: 'Mapping Parameters',
+    children: (
+      <div>
+        <ul>
+          <li>
+            The unselected key will be obmitted during the mapping
```
jfhenon, line 46: omitted

jfhenon (changes requested):

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,230 @@
...
+    return valueConverted
+  }
+
+  /**
+   * @return {Promise} - Promise
+   */
+  async sendData() {
+    return new Promise((resolve, reject) => {
+      try {
+        const maxLength = 100
```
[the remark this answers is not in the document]
victor-tndl: The aim is to let the user the possibility to change it (100 could be de default value), if you agree. It's a WIP feature so it may evolve.

src/client/components/OIbForm/OIbAuthentication.jsx
```diff
@@ -37,32 +39,55 @@ const OIbAuthentication = ({ value, name, onChange, mode, label }) => {
               label="Type"
               onChange={handleChange}
               value={value.type}
-              options={['Basic']}
+              options={['Basic', 'Custom']}
```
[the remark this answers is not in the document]
victor-tndl: I've talk with @burgerni10  and 'custom' emerged from our talk but we can change it.

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,230 @@
...
+  /**
+   * Send the file.
+   * @param {String} filePath - The path of the file
+   * @return {Promise} - The send status
+   */
+  async handleFile(filePath) {
+    // Verify that the file receive is a csv one
+    const regexExp = RegExp('.csv$')
+
+    if (!regexExp.test(filePath)) {
```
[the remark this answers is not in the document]
victor-tndl: The tips would be explained thanks to error message or in the front?

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,230 @@
...
+      this.logger.error(`Invalid file format, file (${filePath} skipped`)
+      return ApiHandler.STATUS.LOGIC_ERROR
+    }
+
+    await this.parseAndMapp(filePath)
```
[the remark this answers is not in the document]
victor-tndl: In my code, the errors are handle in the functions. Each method catch error and log them but maybe I can add a `throw ApiHandler.STATUS.COMMUNICATION_ERROR` for each catch. Do you want it?

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,230 @@
...
+
+    if (this.httpBody.length !== 0) {
+      await this.sendData()
+    }
```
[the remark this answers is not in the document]
victor-tndl: If the body is empty, it comes from the configuration of the user and the file. So I don't think an error is best approach. Maybe a warning but not sure it will be useful.

sendData return a promise, so if the sendRequest method fails it will not return a success.

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,230 @@
...
+   * @return {Promise} - The send status
+   */
+  convertToHttpBody(jsonObject) {
+    try {
+      if (jsonObject.length > 0) {
+        jsonObject.forEach((jsonObj) => {
+          const response = this.convertCSVRowIntoHttpBody(jsonObj)
+          // Test the result of the mapping/convertion before add it in the httpBody
+          if (response) {
+            this.httpBody.push(response)
```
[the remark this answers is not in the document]
victor-tndl: For me a global variable has the good scope but it's not the good solution because it main contains sensitive information.

Moreover, a class variable is so much easier in the comprehension of the code (same variable from the beginning to the end),  

If you really want a local variable I can implement it but in the optic to parse other type of data, it would be possible to do both at the same time ( httpBodyFromCsv and httpBodyFromJson...). However you may implement 2 different methods and different local variable.

src/north/CsvToHttp/CsvToHttp.class.js
victor-tndl: Indeed I use the same call to the method (It comes from the change with OibAuthentication)... I will correct it. Should I even use a switch?

src/client/components/OIbForm/OIbAuthentication.jsx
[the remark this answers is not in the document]
jfhenon: I am not fan but leave the final word to both of you.

src/north/CsvToHttp/CsvToHttp.class.js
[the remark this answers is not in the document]
jfhenon: The easiest is to use the Logger.

src/north/CsvToHttp/CsvToHttp.class.js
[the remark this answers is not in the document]
jfhenon: @burgerni10 We need to document the return values of handlers in a readme to explain how to develop a North/South handler. 
@victor-tndl Yes, return values are important as they indicate to the cache what to do with the file (csv files for your handler).

src/north/CsvToHttp/CsvToHttp.class.js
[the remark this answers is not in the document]
jfhenon: sure

src/north/CsvToHttp/CsvToHttp.class.js
jfhenon: if it is really the same method, I guess it's better to even not use a switch

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,230 @@
...
+
+  /**
+   * @param {Object} jsonObject - jsonObj
+   * @return {Object} - The send status
+   */
+  convertCSVRowIntoHttpBody(jsonObject) {
+    const object = {}
+    this.mapping.forEach((mapping) => {
+      if (jsonObject[mapping.csvField]) {
+        const field = mapping.httpField
```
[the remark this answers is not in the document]
victor-tndl: If I use a log.silly or log.debug, with a 1000 rows it may have 1000 silly/debug log in the journal....

src/north/CsvToHttp/CsvToHttp.class.js
victor-tndl: I see the issue but not sure how to deals with...

victor-tndl:
@burgerni10 @jfhenon What kind of tests do you expect?

src/north/CsvToHttp/CsvToHttp.class.js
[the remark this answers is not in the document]
Rhmaric: Sure, but @jfhenon's comment remains worth considering. Before executing the `convertCSVRowIntoHttpBody` on each row, we can first check that all mapping properties exist in the .csv header fields and log a single message per mispelled or unknown field ?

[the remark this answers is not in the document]
Rhmaric: In silly mode, we can even consider send a message to log.silly with all the available .csv headers

src/north/CsvToHttp/CsvToHttp.class.js
victor-tndl: `if (jsonObject[mapping.csvField])` is false because we want to test the field not the value of the field. So, I rebase the test to the original one

victor-tndl:
@burgerni10 Must I change the test for OiBAuthentication or you prefer do it?

src/north/CsvToHttp/CsvToHttp.class.js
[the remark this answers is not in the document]
victor-tndl: I've done something in order to log available headers in silly mode. However, we should consider to log it as a table (maybe improve the logger service).

burgerni10:
> @burgerni10 Must I change the test for OiBAuthentication or you prefer do it?

The changes introduced a bug referenced in issue #995 , can you check that at the same time you adapt the tests ?

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,218 @@
...
+   * @param {Engine} engine - The Engine
+   * @return {void}
+   */
+  constructor(applicationParameters, engine) {
+    super(applicationParameters, engine)
+
+    const { applicativeHostUrl, requestMethod, bodyMaxLength, authentication, csvDelimiter, mapping, proxy = null } = applicationParameters.CsvToHttp
+
+    // Creation of an entity request with is composed of all necessaries informations
+    this.request = {}
```
jfhenon, line 23: this.request = {
key1: value1,
...}

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,218 @@
...
+
+    return ApiHandler.STATUS.SUCCESS
+  }
+
+  /**
+   * @param {string} filePath - Path of the file to parse
+   * @return {Promise} - the parsed content of the file
+   */
+  async parseCsvFile(filePath) {
+    return new Promise((resolve, reject) => {
```
jfhenon, line 79: Await style?

src/north/CsvToHttp/CsvToHttp.class.js
jfhenon: We tend to avoid using prototype calls. ES6 has usually workarounds.

src/services/request.service.js
```diff
@@ -215,10 +215,15 @@ const sendRequest = async (engine, requestUrl, method, authentication, proxy, da
 
   // Generate authentication header
   const headers = baseHeaders
-  if (authentication && (authentication.type === 'Basic')) {
-    const decryptedPassword = engine.encryptionService.decryptText(authentication.password)
-    const basic = Buffer.from(`${authentication.username}:${decryptedPassword}`).toString('base64')
-    headers.Authorization = `Basic ${basic}`
+  if (authentication) {
```
jfhenon, line 218: should be a "switch" as we will likely have more type in the future.

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,218 @@
...
+      default: return value
+    }
+  }
+
+  /**
+   * @return {Promise} - Promise
+   */
+  async sendData() {
+    return new Promise((resolve, reject) => {
+      try {
```
jfhenon, line 186: Promise vs Async?

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,218 @@
...
+  /**
+   * Convert the value in the selected type (string by default)
+   * @param {Mixed} value - value
+   * @param {string} type - type
+   * @return {Mixed} - The converted value
+   */
+  static convertToCorrectType(value, type) {
+    let valueConverted
+    switch (type) {
+      case 'integer': return parseInt(value, 10) ? parseInt(value, 10) : value
```
jfhenon, line 156: parseInt(value, 10) ?? value

jfhenon: idem timestam/float

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,218 @@
...
+      case 'float': return parseFloat(value) ? parseFloat(value) : value
+      case 'timestamp': return parseInt(value, 10) ? parseInt(value, 10) : value
```
jfhenon, line 158: just want to make sure. If you fail to convert, you just return the initial value? no error? no log?

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,218 @@
...
+      case 'date (ISO)': {
+        valueConverted = new Date(value)
```
jfhenon, line 160: this one is not acceptable. a value global to the scope and reused in several places is a good recipe for bugs. 
please repeat the declaration for each scope

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,218 @@
...
+
+    // Creation of an entity request with is composed of all necessaries informations
+    this.request = {}
+    this.request.host = applicativeHostUrl
+    this.request.method = requestMethod
+    this.request.authentication = authentication
+
+    this.bodyMaxLength = bodyMaxLength
+
+    this.mapping = mapping || {}
```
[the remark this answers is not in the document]
victor-tndl: It comes from Nicolas first review... May I change it?

src/services/request.service.js
victor-tndl: It also comes from another old review

src/north/CsvToHttp/CsvToHttp.class.js
victor-tndl: parseInt('example', 10) return Nan, so I don't this is the best way to solve it. Maybe there is another operator which I don't already know, let me hear if you have one.

src/services/request.service.js
burgerni10: What I meant was to check once if authentication is present as it is now. What means JF I think is that inside this if, a switch(authentication.type) would be better.

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,220 @@
...
+
+      // Parse the csv file into a json object
+      Papa.parse(csvFile, {
+        delimiter: this.csvDelimiter,
+        header: true,
+        dynamicTyping: true,
+        error: (error) => {
+          this.logger.error(error)
+          reject()
+          return []
```
burgerni10, line 93: Is this line necessary with the reject just before ?

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,220 @@
...
+      authenticationField: authentication,
+    }
+
+    this.bodyMaxLength = bodyMaxLength
+
+    this.mapping = mapping || {}
+    this.csvDelimiter = csvDelimiter
+    this.proxy = this.getProxy(proxy)
+
+    this.httpBody = []
```
burgerni10, line 35: httpBody should not be a class variable : you use it like a local variable in handleFile which should be the returned value of the convertToHttpBody function and then should be used as an argument for your sendData function

src/services/request.service.js
victor-tndl: I'll do that, for the default case, should raise an error?

src/north/CsvToHttp/CsvToHttp.class.spec.js
```diff
@@ -0,0 +1,140 @@
...
+  CsvToHttpNorth.httpBody = []
+})
+
+describe('CsvToHttp', () => {
+  it('should be properly initialized', () => {
+    expect(CsvToHttpNorth.canHandleValues).toBeFalsy()
+    expect(CsvToHttpNorth.canHandleFiles).toBeTruthy()
+  })
+
+  it('should properly handle files (not csv one)', async () => {
```
burgerni10, line 37: should properly reject file if type is other than csv

src/north/CsvToHttp/CsvToHttp.class.spec.js
```diff
@@ -0,0 +1,140 @@
...
+    const response = await CsvToHttpNorth.handleFile('../../../tests/csvToHttpTest')
+
+    expect(response).toEqual(ApiHandler.STATUS.LOGIC_ERROR)
+  })
+
+  it('should properly handle files (csv one)', async () => {
```
burgerni10, line 43: should properly handle csv files

src/north/CsvToHttp/CsvToHttp.class.spec.js
```diff
@@ -0,0 +1,140 @@
...
+    const response = await CsvToHttpNorth.handleFile(path.resolve('./tests/csvToHttpTest.csv'))
+
+    expect(response).toEqual(ApiHandler.STATUS.SUCCESS)
+  })
+
+  it('should properly convert values', async () => {
```
burgerni10, line 49: This test should be adapted to the latest changes (it always return objects, with errors if a problem occurs)

src/north/CsvToHttp/CsvToHttp.class.spec.js
```diff
@@ -0,0 +1,140 @@
...
+    expect(typeof response).toEqual(expectedType)
+    expect(response).toEqual(new Date('1998-07-12T21:00:00.000Z'))
+
+    response = CsvToHttp.convertToCorrectType('1998-07-12T21:00:00.000Z', 'short date (yyyy-mm-dd)')
+    expectedType = 'string'
+    expect(typeof response).toEqual(expectedType)
+    expect(response).toEqual('1998-07-12')
+  })
+
+  it('should properly convert wrong values', async () => {
```
burgerni10, line 82: Same than previous test

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,275 @@
...
+    // Logs all the erros
+    if (convertionErrorBuffer.error.length > 0) {
+      this.logger.error(`${convertionErrorBuffer.error.length} convertions error`)
+
+      if (convertionErrorBuffer.error.length > ERROR_PRINT_SIZE) {
+        for (let i = 0; i < ERROR_PRINT_SIZE; i += 1) {
+          this.logger.error(`${convertionErrorBuffer.error[i]}`)
+        }
+      } else {
+        for (let i = 0; i < convertionErrorBuffer.error.length; i += 1) {
```
burgerni10, line 92: forEach may be better here

burgerni10 (changes requested):
Tests should be rework a little bit for CsvToHttp, but otherwise ok for me.
Please check that the fix I did with OIbAuthantication works for you too

victor-tndl:
@burgerni10 Its seems to be good for OiBauthentication. Moreover I've fixed and improved the tests

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,279 @@
...
+    if (convertionErrorBuffer.error.length > 0) {
+      this.logger.error(`${convertionErrorBuffer.error.length} convertions error`)
+
+      if (convertionErrorBuffer.error.length > ERROR_PRINT_SIZE) {
+        for (let i = 0; i < ERROR_PRINT_SIZE; i += 1) {
+          this.logger.error(`${convertionErrorBuffer.error[i]}`)
+        }
+      } else {
+        convertionErrorBuffer.error.forEach((index) => {
+          this.logger.error(`${convertionErrorBuffer.error[index]}`)
```
burgerni10, line 93: This will not work : forEach first parameter is the entity, second parameter is the index. Here you name the entity index which is wrong and use it as an index, which is wrong too. It must be written like this : 
       convertionErrorBuffer.error.forEach((error) => {
          this.logger.error(`${error}`)
        })
        
A test may be usefule to check if the logger is called with the expected error

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,279 @@
...
+    const regexExp = RegExp('.csv$')
+
+    if (!regexExp.test(filePath)) {
+      this.logger.error(`Invalid file format (.csv file expected), file (${filePath} skipped`)
+      return ApiHandler.STATUS.LOGIC_ERROR
+    }
+
+    const csvDataParsed = await this.parseCsvFile(filePath)
+
+    if (csvDataParsed.length === 0) {
```
burgerni10, line 68: a logger (debug level) could be useful here

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,279 @@
...
+        header: true,
+        dynamicTyping: true,
+        error: (error) => {
+          this.logger.error(error)
+          reject()
+        },
+        step: (step) => {
+          csvDataParsed.push(step.data)
+        },
+        complete: () => {
```
burgerni10, line 123: a log (silly level) could be useful here, specifying the filename parsed

src/north/CsvToHttp/CsvToHttp.class.spec.js
```diff
@@ -0,0 +1,165 @@
...
+    expect(typeof response.value).toEqual(expectedType)
+    expect(response.value).toEqual(new Date('1998-07-12T21:00:00.000Z'))
+
+    response = CsvToHttp.convertToCorrectType('1998-07-12T21:00:00.000Z', 'short date (yyyy-mm-dd)')
+    expectedType = 'string'
+    expect(typeof response.value).toEqual(expectedType)
+    expect(response.value).toEqual('1998-07-12')
+  })
+
+  it('should properly convert wrong values', async () => {
```
burgerni10, line 82: should fail to convert wrong values

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  /**
+   * @param {Object} csvFileInJson - json object of the csv file
+   * @return {void}
+   */
+  convertToHttpBody(csvFileInJson) {
+    const httpBody = []
+    // Reset the buffer for error
+    convertionErrorBuffer.reset()
```
burgerni10, line 137: I've just noticed that this variable is declared globally but used locally. Please use the same logic than httpBody : return both from the function and declare another local value in the handleFile function like you did for httpBody

burgerni10: And this way, you won't need a reset function

victor-tndl: The only issue is the use of the variable `convertionErrorBuffer` in the function handleFile(), we should change the response in the function `convertToHttpBody`

burgerni10 (approved):

Issue #1021, opened 2021-01-05
[South - MQTT] - Offer the possibility to establish a persistent connection


burgerni10:
Persistent connections are available in MQTT for QoS 1 & QoS2 : I think we should offer an option to support that as a MQTT client. Here is a quick summary of that : http://www.steves-internet-guide.com/mqtt-clean-sessions-example/

With the node mqtt package, this option is the parameter _clean_ in the connection options.

kukukk:
Based on the documentation it is not enough to disable `clean session` and set QOS > 0 on the subscriber side. The publisher must also connect using QOS > 0. Should it be documented on the UI?
It also requires a persistent client ID on the subscriber side. Should it be configurable by the user or hard-code it to eg. `OIBus`? I think the value is not relevant, but has to be something unique, so the MQTT server can recognize the client. So, in case of multiple OIBus connection to the same MQTT server using the same client ID can cause problems.

burgerni10:
The clean session should be offered only whith QoS1 or QoS2, and a warning (or regular documentation) should tell the user that the clean session will be used only if supported by the broker.

Maybe the client ID can also be reset or just set by the user from the MQTT connector so we can restart and identify the connection better (for debugging for example). If the client ID field is empty, so it can be generated automatically like 'OIBus-<random_number>' with random number large enough to assure a unique client ID.
This has to be documented for the user of course.

In this case, the client ID should be saved in the config file. What do you think about this ?

@jfhenon do you see other use cases / options for this mqtt client ID / session ?

jfhenon:
No but only question is why save to the config file and not SQLite.

burgerni10:
If the user can modify it for connection purposes, it is more a setting than a variable, so it should be stored in the config file I think.
On the other hand, if the client id is generated automatically, it is more a variable, and should be stored in SQLite. It depends how we choose to see the client ID...

kukukk:
Right now we don't have such functionality to make available a field based on another field. If documenting the situation when the clean session will have effect is not enough, I will ask @attilasuto to help modifying the checkbox component to have an enabled/visible function which can use the value of an another field.

The default value for client ID is `clientId: 'mqttjs_' + Math.random().toString(16).substr(2, 8)`. If client ID is not specified I can use the same logic to generate it, but replace `mqttjs_` with `OIBus-`. It would look like: `OIBus-b403429b`

If client ID can be specified on the UI, I think it should be handled the same way as other parameters.

PR #1047, opened 2021-01-18
South mqtt persistent connection#1021

kukukk:
Implement #1021

### DIFF
--- a/src/south/MQTT/MQTT.class.js
+++ b/src/south/MQTT/MQTT.class.js
@@ -21,6 +21,8 @@ class MQTT extends ProtocolHandler {
       username,
       password,
       qos,
+      persistent,
+      clientId,
       dataArrayPath,
       valuePath,
       nodeIdPath,
@@ -41,6 +43,8 @@ class MQTT extends ProtocolHandler {
     this.username = username
     this.password = Buffer.from(this.encryptionService.decryptText(password))
     this.qos = qos
+    this.persistent = persistent
+    this.clientId = clientId || `OIBus-${Math.random().toString(16).substr(2, 8)}`
     this.dataArrayPath = dataArrayPath
     this.valuePath = valuePath
     this.nodeIdPath = nodeIdPath
@@ -61,6 +65,8 @@ class MQTT extends ProtocolHandler {
     const options = {
       username: this.username,
       password: this.password,
+      clientId: this.clientId,
+      clean: !this.persistent,
     }
     this.client = mqtt.connect(this.url, options)
     this.client.on('error', this.handleConnectError.bind(this))
--- a/src/client/South/Form/__snapshots__/SouthForm.spec.jsx.snap
+++ b/src/client/South/Form/__snapshots__/SouthForm.spec.jsx.snap
@@ -1212,6 +1212,18 @@ exports[`SouthForm check SouthForm with dataSource: MQTTServer 1`] = `
                     </li>
                     <li>
                       <b>
+                        Persistent:
+                      </b>
+                      In this mode the broker will store subscription information, and undelivered messages for the client. With a non persistent connection the broker does not store any subscription information or undelivered messages for the client. For this option to take effect QoS must be set to 1 or 2.
+                    </li>
+                    <li>
+                      <b>
+                        Client ID:
+                      </b>
+                      In order for the broker to store session information for a client a client id must be used.
+                    </li>
+                    <li>
+                      <b>
                         Username:
                       </b>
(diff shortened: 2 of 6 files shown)


src/client/components/OIbForm/OIbCheckBox.jsx
```diff
@@ -13,6 +16,19 @@ const OIbCheckBox = ({ label, value, name, onChange, defaultValue, switchButton
     const { checked } = target
     onChange(name, checked, null)
   }
+  if (dependency) {
+    // check if dependency exists, and compare the value
+    // to the dependencyValues list to enable/disable component
+    const dependecyValue = objectPath.get(newConfig, dependency)
```
jfhenon, line 22: spelling mistake -> dependencyValue

src/south/MQTT/MQTT.class.js
```diff
@@ -41,6 +43,8 @@ class MQTT extends ProtocolHandler {
     this.username = username
     this.password = Buffer.from(this.encryptionService.decryptText(password))
     this.qos = qos
+    this.persistent = persistent
+    this.clientId = clientId || `OIBus-${Math.random().toString(16).substr(2, 8)}`
```
jfhenon, line 47: our nodejs supports the ?? notation..

src/client/components/OIbForm/OIbCheckBox.jsx
```diff
@@ -13,6 +16,19 @@ const OIbCheckBox = ({ label, value, name, onChange, defaultValue, switchButton
     const { checked } = target
     onChange(name, checked, null)
   }
+  if (dependency) {
+    // check if dependency exists, and compare the value
```
jfhenon, line 20: is this test necessary as the hook UseEffect is controlled by [dependecyValue]?

src/client/components/OIbForm/OIbCheckBox.jsx
```diff
@@ -1,8 +1,11 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { FormGroup, Label, Input, CustomInput } from 'reactstrap'
+import objectPath from 'object-path'
+import { ConfigContext } from '../../context/configContext.jsx'
 
-const OIbCheckBox = ({ label, value, name, onChange, defaultValue, switchButton }) => {
+const OIbCheckBox = ({ label, value, name, onChange, defaultValue, switchButton, dependency, dependencyValues }) => {
```
jfhenon, line 7: I like the concept of dependencies but I guess it should not be specific to only check boxes. This adds some complexity to OIBCheckBox that should not be handled at this level. 

I tend to think that this should be implemented at the Form level for every Component and in a different PR. Actually, the option should allow to disable or hide the component.

jfhenon (changes requested):
Please see my comments

src/south/MQTT/MQTT.class.js
kukukk: **??** is the nullish coalescing operator which will only use the value from the right side when the value on left side is null or undefined. It means that if `clientId` is an empty string, it will use the empty string which is not valid for MQTT.
Using **||** will ensure that `this.clientId` is not an empty string, because it will use the value from the right side if the value on the left side is falsy.

jfhenon: ok, my mistake.

src/client/components/OIbForm/OIbCheckBox.jsx
attilasuto: Spelling mistake fixed.

src/client/components/OIbForm/OIbCheckBox.jsx
attilasuto: The "if (dependency)" test is needed because we only need to check the dependencyValue if we have a dependency (as path) defined, and the hook is controlled by  dependencyValue which exists based on the path.

src/client/components/OIbForm/OIbCheckBox.jsx
attilasuto: You're right, we can handle on Form level for all kind of components, do we want to change this on this PR or should we create a separate ticket and handle in a dedicated PR? Do we need to change anything related to this on this PR?

jfhenon: I would prefer this to be separate PR. For now, we can remove the dependencies of OIBCheckBox so we can close this PR and reopen a new one.

attilasuto: The dependency of OIBCheckBox was removed.

Issue #981, opened 2020-12-08
Refactor the request service


kukukk:
Instead of the single `sendRequest()` public method we should have separate methods to post a file, a json content or compressed json content and handle the content based on the called method instead of the custom header fields passed to `sendRequest()`

Maybe we should also separate the node-fetch and axios implementations into separate classes and use a factory to instantiate the right one based on the stack configured in Enginge.

burgerni10:
In the refactoring process, it would be nice to add some comments on the sendRequest functions about the ApiHandler.STATUS

PR #1048, opened 2021-01-18
Refactor request service#981

kukukk:
Implement #981


src/engine/AliveSignal.class.js
```diff
@@ -60,7 +60,7 @@ class AliveSignal {
     try {
       const data = JSON.stringify(status)
       const headers = { 'Content-Type': 'application/json' }
-      await this.engine.sendRequest(this.host, 'POST', this.authentication, this.proxy, data, headers)
+      await this.engine.requestService.send(this.host, 'POST', this.authentication, this.proxy, data, headers)
```
jfhenon, line 63: maybe we should have a method proposed for every North handlers so it could be called this.send or this.HttpSend. Also, to make it more readable and also propose defaults, I propose to pass an object
{ url, mode = Post, authentication etc... }

src/services/request/RequestFactory.class.js
```diff
@@ -0,0 +1,15 @@
+const AxiosRequest = require('./AxiosRequest.class')
+const FetchRequest = require('./FetchRequest.class')
+
+class RequestFactory {
```
jfhenon, line 4: I believe a class with only a single static method should not be a class. 
We should just export a function or move this logic directly in the North generic class.

jfhenon (changes requested):
Very good for the overall look of the code/comment.

However I find it a little heavy for the developer of a "north" developper. 
Let me propose a few things in my different comments.

src/services/request/RequestFactory.class.js
kukukk: The reuqest service is also used in AliveSignal, so we can't make it North specific. I can export a function or move the static create() method into the BaseRequest class. Which one do you prefer?

jfhenon: No preference.

src/engine/AliveSignal.class.js
kukukk: I wanted a minimal impact on other components, but I was already thinking to make some 'alias' methods, like `postJsonValue`, `postFile`, etc. where the parameters are only specific for the caller.
Since AliveSignal also uses the RequestService, I would implement these methods in BaseRequest, but I also can create sendValues() and sendFile() in ProtocolHandler as an additional abstraction level for North developers.
What do you think?

jfhenon (approved):

burgerni10 (approved):

Issue #1076, opened 2021-02-02
[North-CsvToHttp] Use variable in input's field


victor-tndl:
This feature aim to let the use the choice to use variable in input's field. He'll be able to concatenate string using template string. This feature is related to #1005.

For example:
A North "CsvToHttp" receives a CSV file from a South. Inside, we have 2 headers called "header1" (its value is "example1" for each row) and "header2" (its value is "example2" for each row)
The user wants a custom output which is the concatenation of both headers.  The user will set the field "Header in csv file" to  "${header1} and ${header1}".

In this case, the result will be "example1 and example2".

In order to implement this, the code may be something like this:
```js
public function convert(value){
     result = eval('`'+value+'`')
     return result
}
```

Let me know if everything is ok for you (@burgerni10 @jfhenon ).

jfhenon:
I would not approve a PR using eval() which causes security concerns. If you need to concatenate strings, you have several packages that could be used instead. 
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/eval

victor-tndl:
I was pretty sur that it would not have been approuved. I'll try to find another way to use variable.

However, is aim fo this feature ok four you?

burgerni10:
I agree with JF, eval should be avoided.

I think the feature can be useful. You can start thinking more about it (without eval of course !).

We must discuss internally a little bit more about the way the variables are defined : in SqlDbToFile, we use @variable but something code-like as you propose may be better. @jfhenon what do you think ?

jfhenon:
our MQTT handler (and others) uses sprintf-js. This should be good for the use case above.
JF

victor-tndl:
I already used regex in my solution

PR #1080, opened 2021-02-03
Use template strings in input (#1076)

victor-tndl:


src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -174,12 +267,47 @@ class CsvToHttp extends ApiHandler {
           object.error.push(`Header "${mapping.httpField}": ${response.error}`)
         }
         object.value[field] = response.value
+      } else if (mapping.csvField.match(REGEX_CONTAIN_VARIABLE_STRING)) {
+        // split the input into a array of string
+        // "test ${value}" => [ "test ", ${value}]
+        const csvFieldSplit = mapping.csvField.split(REGEX_SPLIT_TEMPLATE_STRING).filter(Boolean)
```
burgerni10, line 273: filter is not useful here

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -174,12 +267,47 @@ class CsvToHttp extends ApiHandler {
...
+        const field = mapping.httpField
+        csvFieldSplit.forEach((element) => {
+          if (element.match(REGEX_MATCH_VARIABLE_STRING)) {
```
burgerni10, line 276: comment this (same comment than the one at the top of the file) to explain what you get here

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -174,12 +267,47 @@ class CsvToHttp extends ApiHandler {
...
+            const headerToGet = element.match(REGEX_GET_VARIABLE)
```
burgerni10, line 277: comment this (same comment than the one at the top of the file) to explain what you get here

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -174,12 +267,47 @@ class CsvToHttp extends ApiHandler {
...
 
+  /**
+   * @param {Mixed} object - object
+   * @param {Mixed} value - value
+   * @return {Mixed} - The converted value
+   */
+  static insertValueInObject(object, value) {
+    if (object) {
+      let response = object
+      response += value
```
burgerni10, line 305: If I understand correctly, in case of aggregated field, the value is added to the previous one ? 
It may need a separator no ?
You may improve the doc of this function for better understanding

src/north/CsvToHttp/CsvToHttp.schema.jsx
```diff
@@ -81,6 +81,22 @@ schema.form = {
...
+        </p>
+        <p>
+          <b>Example 2: </b>
+          With the same csv file, it is possible to specify the name for each header in the http request using template string
+          (references to other header as a variable). The syntax is the js one (such as: $
+          {'{value}'}
+          ).It is possible to have in the request&lsquo;s body a field called header1AndHeader2 its value is the concatenation of both
+          header&lsquo;s value.
+          The &quot;Header in csv file&quot; must be filled with the value: $
+          {'{header1} and {header2}'}
```
burgerni10, line 97: I think that the examples here are a little bit complicated for a user who starts using this connector. Can you try to improve it please ?

src/north/CsvToHttp/CsvToHttp.class.js
victor-tndl: I think it's useful because in the REGEX we have a "or" logic  which means that without it we can have empty string sent

burgerni10: My bad, ok for this one !

jfhenon:
One question, as we are using sprintf as a library. Why don't we use the native syntax (with % like the example below) instead of parsing the string to find the ${}. I feel we add complexity in the code for no real gain for the end user? 
````
var row = {
    id: 'Dolly',
}
sprintf('Hello %(id)s', row) // Hello Dolly
````

File: src/north/CsvToHttp/CsvToHttp.class.js [hunk 6]
```diff
@@ -158,14 +159,106 @@ class CsvToHttp extends ApiHandler {
...
+  /**
+   * Test if the field is valid thanks to all headers
+   * @param {Mixed} allHeaders - available headears in the csvFile
+   * @param {Mixed} field - field to test if it matchs with the available headears
+   * @return {Boolean} - Return a bool
+   */
+  static isHeaderValid(allHeaders, field) {
+    if (field.match(REGEX_CONTAIN_VARIABLE_STRING)) {
+      const csvFieldSplit = field.split(REGEX_SPLIT_TEMPLATE_STRING).filter(Boolean)
+      csvFieldSplit.every((element) => {
```
burgerni10, line 234: You must add "return" at the start of the line. Otherwise, this test (the 'every' loop) never returns true or false. it then goes on to return false on line 248 since the field is not found on the allHearders tab

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -174,12 +267,50 @@ class CsvToHttp extends ApiHandler {
...
+      } else if (mapping.csvField.match(REGEX_CONTAIN_VARIABLE_STRING)) {
+        // split the input into a array of string
+        // "test ${value}" => [ "test ", ${value}]
+        const csvFieldSplit = mapping.csvField.split(REGEX_SPLIT_TEMPLATE_STRING).filter(Boolean)
+        const field = mapping.httpField
+        csvFieldSplit.forEach((element) => {
+          // match if the string starts with: ${...}
+          if (element.match(REGEX_MATCH_VARIABLE_STRING)) {
+            const headerToGet = element.match(REGEX_GET_VARIABLE)
+            headerToGet.forEach((header) => {
```
burgerni10, line 279: Correct me if I'm wrong, but from my understanding : headerToGet is the value extracted from element. If element is ${test}, so headerToGet will be [ 'test' ].

My point is : I don't see the case where you need to loop inside an array. It must always be one item inside the resulting array (because of the previous regex filtering). It may be clearer to remove this forEach. 
Instead, I would do something like this : 
const header element.match(REGEX_GET_VARIABLE)[0]
And then something like : if(header) { do stuff }

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -174,12 +267,50 @@ class CsvToHttp extends ApiHandler {
...
   }
 
+  /**
+   * It returns the concatenation of value with the previous object
+   * If the oject is empty it return the value sent
+   * @param {Mixed} object - object
+   * @param {Mixed} value - value
+   * @return {Mixed} - The converted value
+   */
+  static insertValueInObject(object, value) {
```
burgerni10, line 305: To make it clearer, I would rename it 'currentJsonValue'. Don't forget to update the jsdoc.

src/north/CsvToHttp/CsvToHttp.class.js
victor-tndl: If we have the input "${test} and ${test2}" the regex will match with both header. We'll get an array of both value. 
Here, we ensure that the headers "test" and "test2" are present in the CSV file. However, we notice the user that his input isn't valid.

src/north/CsvToHttp/CsvToHttp.class.js
burgerni10: And rename value to 'valueToAdd'

src/north/CsvToHttp/CsvToHttp.class.js
burgerni10: const csvFieldSplit = mapping.csvField.split(REGEX_SPLIT_TEMPLATE_STRING).filter(Boolean) should split "${test} and ${test2}" to ["${test}", " and ", "${test2}"] (line 273)

You then make a forEach on this array. So you parse individually each element. That's why I say that on line 278, you will always have only one item. So the forEach of line 279 is useless.

Am I right ?

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -174,12 +267,50 @@ class CsvToHttp extends ApiHandler {
...
+          // match if the string starts with: ${...}
+          if (element.match(REGEX_MATCH_VARIABLE_STRING)) {
+            const headerToGet = element.match(REGEX_GET_VARIABLE)
+            headerToGet.forEach((header) => {
+              const response = CsvToHttp.convertToCorrectType(csvRowInJson[header], mapping.type)
+              if (response.error) {
+                object.error.push(`Header "${mapping.httpField}": ${response.error}`)
+              }
+              if (response.value) {
+                object.value[field] = CsvToHttp.insertValueInObject(object.value[field], response.value)
```
burgerni10, line 285: what if we add an integer to a previous value which was a date (or other cases like string + integer, etc) ? Should you not test beforehand that the type are compatible and trigger a warning/error log otherwise ?

victor-tndl: My mistake it handles only string value (I'll change the conversion type)

src/north/CsvToHttp/CsvToHttp.schema.jsx
```diff
@@ -81,6 +81,25 @@ schema.form = {
...
+        <p>
+          <b>Example 1: </b>
+          If the CSV file contains 2 headers called header1 and header2, it is possible to specify the name for each field in the
+          http request. It&lsquo;s possible to have in the body of the request a field called &quot;AfterMapping&quot;
+          which contains the value of the header1 for each row. We can do the same for each header present in the CSV file
+        </p>
+        <p>
+          <b>Example 2: </b>
+          It is possible to specify the name for each header in the http request using template string
+          (references to other header as a variable). Thanks to it we can reference header present in the CSV file as variable.
```
burgerni10, line 93: to other headers...

we can reference headeers prensent...
...as variables.

src/north/CsvToHttp/CsvToHttp.schema.jsx
```diff
@@ -81,6 +81,25 @@ schema.form = {
...
+          For it, the syntax to use is the javascript one (such as: $
+          {'{name of your header}'}
+          ).
+          With the same CSV file, it is possible to have in the request&lsquo;s body a field called header1AndHeader2 its value is the
+          concatenation of both header&lsquo;s value.
+          The &quot;Header in CSV file&quot; must be filled with the value: $
+          {'{header1} and {header2}'}
+          .
```
burgerni10, line 101: Can you add a small sentence like : if a row has 'value1' in the column 'header1' and 'value2' in the column 'header2', the resulting value will be 'value1 and value2'

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -276,15 +278,20 @@ class CsvToHttp extends ApiHandler {
           // match if the string starts with: ${...}
           if (element.match(REGEX_MATCH_VARIABLE_STRING)) {
             const headerToGet = element.match(REGEX_GET_VARIABLE)
-            headerToGet.forEach((header) => {
-              const response = CsvToHttp.convertToCorrectType(csvRowInJson[header], 'string')
+            // The regex must match with only one value and return an array with one element
+            if (headerToGet.length !== 1) {
+              object.error.push(`Regex doesn't match only with one value (tried element: ${element})`)
+            }
```
burgerni10, line 284: Next if must be 'else if' instead. Indeed : if headerToGet.length == 0, you push the error, but then the next if is executed.

burgerni10 (changes requested):

PR #1127, opened 2021-02-28
Aggregates

jfhenon:
These modifications add ability for OPCUA to manage aggregates. 
I was not able to test them on Prosys and WinCC (apparently lack of support)
Please check the code. 

Please note that the HDAAgent was modified for aggregate support in bulk mode. We don't yet have a use case to add this in runtime mode.


src/south/OPCUA_HA/OPCUA_HA.class.js
```diff
@@ -226,6 +249,49 @@ class OPCUA_HA extends ProtocolHandler {
...
+          case 'Minute':
+            options.processingInterval = 1000 * 60
+            break
+          case 'Hour':
+            options.processingInterval = 1000 * 3600
+            break
+          case 'Day':
+            options.processingInterval = 1000 * 3600 * 24
+            break
+          case 'None':
```
kukukk, line 271: Shouldn't we use `0` as default?
```Interval between returned Aggregate values. The value 0 indicates that there is no ProcessingInterval defined.```
From this link: https://reference.opcfoundation.org/v104/Core/docs/Part11/6.4.4/

jfhenon: I believe we should make mandatory to add a resamplingInterval but not take 0 as a default 
0, as you point out rightly, has a specific meaning (one aggregated value from start to end date) so I would not use it as a default. 
In fact, I don't think we should use it with OIBus as it break the logic with lastCompletedAt.

burgerni10 (changes requested):
A warning just on top of the scan group to specify the aggregate functions need to be tested may be good.
Otherwise it's ok !

burgerni10:
@jfhenon I've rebased your branch to fix conflicts and I added minor ui changes. 
LGTM though, I let you review if needed and rebase and merge this branch

PR #1129, opened 2021-03-02
feat(south-modbus): Modbus address offset (ON TOP OF #1128)

Lulu300:
This feature allows to define an offset on the addresses of the points to be requested in order to be compatible with the JBus protocol. When using JBus, all the addresses are offset by -1, the addition of this new field avoids having to increment the addresses of all the points manually.

### DIFF
--- a/src/south/Modbus/Modbus.schema.jsx
+++ b/src/south/Modbus/Modbus.schema.jsx
@@ -57,9 +57,9 @@ schema.form = {
     type: 'OIbSelect',
     md: 2,
     newRow: false,
-    options: [0, -1],
+    options: ['Modbus', 'JBus'],
     label: 'Address Offset',
-    defaultValue: 0,
+    defaultValue: 'Modbus',
     help: <div>Address offset to be applied for all points during requests (0 for the traditionnal Modbus protocol and 1 when using JBus)</div>,
   },
 }
--- a/src/south/Modbus/Modbus.schema.jsx
+++ b/src/south/Modbus/Modbus.schema.jsx
@@ -53,6 +53,15 @@ schema.form = {
     defaultValue: 1,
     help: <div>Slave ID of the Modbus source</div>,
   },
+  addressOffset: {
+    type: 'OIbSelect',
+    md: 2,
+    newRow: false,
+    options: [0, -1],
+    label: 'Address Offset',
+    defaultValue: 0,
+    help: <div>Address offset to be applied for all points during requests (0 for the traditionnal Modbus protocol and 1 when using JBus)</div>,
+  },
 }
 
 schema.points = {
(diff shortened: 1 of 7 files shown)


src/south/Modbus/Modbus.schema.jsx
```diff
@@ -53,6 +53,15 @@ schema.form = {
     defaultValue: 1,
     help: <div>Slave ID of the Modbus source</div>,
   },
+  addressOffset: {
+    type: 'OIbSelect',
+    md: 2,
+    newRow: false,
+    options: [0, -1],
```
burgerni10, line 60: I don't know much about the use cases for the offset, but if it's only for Modbus / Jbus, the select should display these options instead of the integer. On the backend, depending of the value (Modbus or JBus) the right offset (0 or -1) is used. Default setting would be of course Modbus (instead of 0). What do you think ?

burgerni10 (commented):
I rebased your branch so it is up to date with the master branch. I updated your commits to reflect the conventional commit format we use. See https://www.conventionalcommits.org/en/v1.0.0/ for more info. It helps us to automate our release changelogs.

src/south/Modbus/Modbus.schema.jsx
Lulu300: I think it'll be only Modbus and JBus, you're right I'll make the changes. I also have a question, is it possible in the options field to provide key value instead of only value (Like HTML5 Select Form) ?

Lulu300:
> 
> 
> I rebased your branch so it is up to date with the master branch. I updated your commits to reflect the conventional commit format we use. See https://www.conventionalcommits.org/en/v1.0.0/ for more info. It helps us to automate our release changelogs.

I'll be careful to respect this convention for future commits

src/south/Modbus/Modbus.schema.jsx
burgerni10: Right now, it is not possible : the options are an array of values directly used as it is by the OibSelect component. That's why a small process in the class file may be used to convert JBus to "-1" and Modbus to "0"

burgerni10:
Hi @Lulu300, thank you for your work, your PR seems fine now :)

Before accepting it, I was wondering about JBus / Modbus : from what I understand, JBus  addresses from 0 instead of 1 for Modbus, hence the offset.

My question is : does a user directly get the JBus address from the PLC (in this case the offset is confusing because it will be the correct address right away) or does a JBus PLC give the Modbus address to the user and in this case the offset make sense ?

Lulu300:
In fact it depends on the PLC manufacturer but usually they provide the user with the Modbus address, hence we need the offset to make the configuration in OIBus look similar to what's inside the PLC documentation.

burgerni10:
Ok, thanks !

PR #1131, opened 2021-03-04
 feat(south-modbus): Add possibility to specify the data type of a point (ON TOP OFF #1129)

Lulu300:
This feature allows you to define the type of values to be read. By default with the Modbus protocol, the values of the HoldingRegister and InputRegister registers are stored as UInt16. But it is more and more common to see Modbus PLCs storing these values on 32 bits (signed or unsigned), and sometimes these values can be decimal numbers and not integers. In rarer cases the values may even be stored on 64 bits.

### DIFF
--- a/src/south/Modbus/Modbus.class.js
+++ b/src/south/Modbus/Modbus.class.js
@@ -5,21 +5,6 @@ const { getOptimizedScanModes } = require('./config/getOptimizedConfig')
 const ProtocolHandler = require('../ProtocolHandler.class')
 
 /**
- * Read point value according to it's data type (UInt16, UInt32, etc)
- * @param {Object} response Response of the modbus request
- * @param {Object} point The point to read
- * @param {Number} position Position of the point in the response
- * @return {Number} Value stored at the specified address
- */
-const readRegisterValue = (response, point, position) => {
-  if (response.body.constructor.name === 'ReadCoilsResponseBody' || response.body.constructor.name === 'ReadDiscreteInputsResponseBody') {
-    return response.body.valuesAsArray[position]
-  }
-  const funcName = `read${point.dataType}BE`
-  return response.body.valuesAsBuffer[funcName](position * 2)
-}
-
-/**
  * Class Modbus - Provides instruction for Modbus client connection
  */
 class Modbus extends ProtocolHandler {
@@ -63,6 +48,24 @@ class Modbus extends ProtocolHandler {
   }
 
   /**
+   * Read point value according to it's data type (UInt16, UInt32, etc)
+   * @param {Object} response Response of the modbus request
+   * @param {Object} point The point to read
+   * @param {Number} position Position of the point in the response
+   * @return {Number} Value stored at the specified address
+   */
+  readRegisterValue(response, point, position) {
+    if (response.body.constructor.name === 'ReadCoilsResponseBody' || response.body.constructor.name === 'ReadDiscreteInputsResponseBody') {
+      return response.body.valuesAsArray[position]
+    }
+    const endianness = this.dataSource.Modbus.endianness === 'Big Endian' ? 'BE' : 'LE'
+    const funcName = `read${point.dataType}${endianness}`
+    /* Here, the position must be multiplied by 2 because the jsmodbus library is set to read addresses values on 16 bits (2 bytes),
+      but in the valuesAsBuffer field each cell of the array contains 8 bits (1 byte) */
+    return response.body.valuesAsBuffer[funcName](position * 2)
+  }
+
+  /**
    * Dynamically call the right function based on the given name
    * @param {String} funcName - Name of the function to run
    * @param {Object} infos - Information about the group of addresses (first address of the group, size)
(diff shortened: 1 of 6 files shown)


src/south/Modbus/Modbus.class.js
```diff
@@ -4,6 +4,21 @@ const net = require('net')
...
+ * @param {Object} response Response of the modbus request
+ * @param {Object} point The point to read
+ * @param {Number} position Position of the point in the response
+ * @return {Number} Value stored at the specified address
+ */
+const readRegisterValue = (response, point, position) => {
+  if (response.body.constructor.name === 'ReadCoilsResponseBody' || response.body.constructor.name === 'ReadDiscreteInputsResponseBody') {
+    return response.body.valuesAsArray[position]
+  }
+  const funcName = `read${point.dataType}BE`
```
burgerni10, line 18: If I follow our discussions, the 'BE' part will be another parameters to specify Big or Little endian ?

src/south/Modbus/Modbus.class.js
```diff
@@ -4,6 +4,21 @@ const net = require('net')
...
+  return response.body.valuesAsBuffer[funcName](position * 2)
```
burgerni10, line 19: Not sure to understand the 'position*2' part, can you tell me more about it ?
A comment on that in the code may be useful

src/south/Modbus/Modbus.schema.jsx
```diff
@@ -78,6 +78,20 @@ schema.points = {
...
+    newRow: false,
+    options: ['UInt16', 'Int16', 'UInt32', 'Int32', 'UInt64', 'Int64', 'Float', 'Double'],
+    label: 'Data type',
+    defaultValue: 'Uint16',
+    help: (
+      <div>
+        In the case of a value from a holdingRegister or an inputRegister, what is the type of the data
+        (Is it an integer or a decimal number, on how many bits the value is stored and is the value signed or not.).
+        By default for these registers the value is stored on a UInt16
+      </div>
```
burgerni10, line 92: I would rephrase this help like that :
`HoldingRegisters and inputRegisters can have one of the above types. Default type is UInt16. This field does not apply for coils and discreteInputs.`

burgerni10 (changes requested):
I think the UI can be improved :
image

I would put the fields in the following order : pointId, address, ModbusType, dataType and scanMode, with more space for pointId, address and scanMode so we can see right away the content of these fields. Can you adapt it please ?

src/south/Modbus/Modbus.class.js
Lulu300: > 
> 
> I think the UI can be improved :
> image
> 
> I would put the fields in the following order : pointId, address, ModbusType, dataType and scanMode, with more space for pointId, address and scanMode so we can see right away the content of these fields. Can you adapt it please ?

i change the order of the fields, but when I add the "md" parameters on the differents fields of the points schema, the fields ignore this value and stay at the same size. Do you have an idea why ?

src/south/Modbus/Modbus.schema.jsx
Lulu300: Yes, I agree, it's much clearer that way. I'll make the change

src/south/Modbus/Modbus.class.js
Lulu300: Yes, I think I'll add an other OIBSelect field to choose between BE and LE

src/south/Modbus/Modbus.class.js
Lulu300: By default, the jsmodbus library you use reads addresses values in Big Endian and unsigned int 16 bits. So in the response of the query, in the valuesAsArray field that you used to retrieve the value of the point, each index of the array corresponds to 16 bits. On the other hand in the valuesAsBuffer field, each index of the bufer corresponds to 1 byte or 8 bits. If we take the example where it is the value which is at the 5th index of the array of the valuesAsArray field, it will correspond to the index 10 of the valuesAsBuffer field, and instead of reading only one cell of the array you have to read 2, 4 or 8 according to the size of the value to retrieve (Int16, Int32, Int64). I will try to simplify my explanation and put it as a comment in the code, but is it clearer for you already?

Lulu300:
> 
> 
> I think the UI can be improved :
> image
> 
> I would put the fields in the following order : pointId, address, ModbusType, dataType and scanMode, with more space for pointId, address and scanMode so we can see right away the content of these fields. Can you adapt it please ?

I just saw that @jfhenon open an issue to add more customization on the OITable component, should I wait these changes to be added before making my modification on the UI ?

src/south/Modbus/Modbus.class.js
burgerni10: It's clearer, thanks. I missed that point in the jsmodbus library. Thank you for having fixed that !

My tests were kind of trivial, and following your answer, I think it must be improved to take this into account. I can improve it later, but if you can do it now, it would be perfect. !

burgerni10:
> > I think the UI can be improved :
> > image
> > I would put the fields in the following order : pointId, address, ModbusType, dataType and scanMode, with more space for pointId, address and scanMode so we can see right away the content of these fields. Can you adapt it please ?
> 
> I just saw that @jfhenon open an issue to add more customization on the OITable component, should I wait these changes to be added before making my modification on the UI ?

We'll work on the issue #1139 to improve the UI. 

In the meantime, can you put your "help" in the main modbus settings page ? The list page will be lighter this way. I will put back the help in the list page when the issue will be solved.

burgerni10 (approved):

PR #1133, opened 2021-03-04
ADS/develop

burgerni10:


src/services/validation.service.js
```diff
@@ -4,6 +4,8 @@ import path from 'path'
 // eslint-disable-next-line max-len
 const ipv4 = /^\s*((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\s*$/
 // eslint-disable-next-line max-len
+const adsNetId = /^\s*((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\s*$/
+// eslint-disable-next-line max-len
```
jfhenon, line 8: should we put "driver" dependant validation in this validation.service.js supposed to be more generic?

src/south/ADS/ADS.class.js
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    await nodesToRead.forEach((node) => {
+      this.client.readSymbol(node.address)
+        .then((nodeResult) => {
+          this.addValues([
+            {
+              pointId: node.pointId,
+              timestamp: new Date().toISOString(),
+              data: { value: JSON.stringify(nodeResult) },
```
jfhenon, line 52: timestamp is never given by ADS source???

src/south/ADS/ADS.class.js
```diff
@@ -0,0 +1,99 @@
...
+   */
+  async connect() {
+    await super.connect()
+    this.client = new ads.Client({
+      targetAmsNetId: this.netId, // example: 192.168.1.120.1.1
+      targetAdsPort: this.port, // example: 851
+    })
+
+    this.client.connect()
+      .then((res) => {
```
jfhenon, line 74: should we "await this.client.connect"?

src/south/ADS/ADS.class.js
burgerni10: From what I understand, we can get the date with subscription (to be implemented with listen mode later maybe).

Otherwise, we get the data at the date we request it, so this timestamp is correct. The timestamp is not returned by the payload.

src/services/validation.service.js
burgerni10: For now I let it here : we should first discuss how to manage specific validation : do we put it in a 'local' validation service in the associated connector ?

jfhenon (approved):

Issue #1078, opened 2021-02-03
[South] - handle historians protocols with startDate and endDate


burgerni10:
Add a canHandleHistory=true; property for South protocols that can handle historians (like OPC-HDA, OPCUA, SQLDbToFile...)

For protocols that can handle historians, use a startDate and endDate to specify the timeframe of the requested data.

In run mode (the existing mode), when a request is trigger by a scan mode, startDate replaces lastCompletedAt and endDate is always now(). After the request, startDate=endDate.
In this mode, startDate can be set from the south protocol UI. Its value cannot be empty. A new timedatepicker component must be added for the UI.

burgerni10:
A migration for lastCompletedAt may be necessary here.

kukukk:
Right now the logic is the following:
 - first the database is checked for lastCompletedAt. If found, it will be used
 - if not found, startTime is checked in the config. If preset it will be used as lastCompletedAt
 - if not found new Date() will be used

It means that if lastCompletedAt is present in the database startDate/endDate from the config file won't be taken into account.
If I change the priority to check first startDate/endDate from the config it will use these values on the next start instead of the lastCompletedAt stored in the dabatase.
If both startDate and endDate is specified, when reaching endDate we should stop reading?

marouanehassanioptimistik:
if we want to do this cleanly we need to have separate flows for realtime and historic. My guess is that if we try to mix both the logic will be flawed and the tool not as flexible

burgerni10:
Following Marouane comment :
When a request is trigger in "run mode" : same as usual. LastCompletedDate should be displayed on the front though. It can be used to track the run mode with that. See issue #912. When an error occurs, it can still be useful to edit the LastCompletedDate from the UI (with a datetimepicker).

When a request is trigger in bulk : see issue #1079.

@jfhenon do you agree with that following your recent tests ?

burgerni10:
To sum it up, for now canHandleHistory=true must be set in historian-supported protocols (SqlDbToFile, OPC-HDA, OPCUA-HA).

LastCompletedDate must be displayed on the connector UI. For multi-LastCompletedDate protocols (OPCUA-HA with Scan Groups for example), each scan group should display its own LastCompletedDate.

start/endDate will be implemented in bulk mode (see issue #1079) and the process may be merged later if necessary with 'run mode'.

kukukk:
We had a discussion with @attilasuto about what should be the better logic and came up with the following:
[...]

PR #1144, opened 2021-03-10
Show last completed at on UI#1078

attilasuto:
#1078


burgerni10:
Following our internal discussions, a live screen for each south connector will be a good place to display these realtime readonly information (and to separate them from writable settings). These screens will be improved later with other features, like a “web console” to display the connector activity : JF did a poc similar to this, but it’s an improvement for another time.

JF suggestion : create a screen for each protocol (for example: http://localhost:2223/south/myopcua/live) where we will put live informations. This url will be accessible with a button in the list of South (next to the settings button) and a button in the protocol itself (left of the "Points" or "File" button).

burgerni10 (changes requested):
Following our internal discussions, a live screen for each south connector will be a good place to display these realtime readonly information (and to separate them from writable settings). These screens will be improved later with other features, like a “web console” to display the connector activity : JF did a poc similar to this, but it’s an improvement for another time.

JF suggestion : create a screen for each protocol (for example: http://localhost:2223/south/myopcua/live) where we will put live informations. This url will be accessible with a button in the list of South (next to the settings button) and a button in the protocol itself (left of the "Points" or "File" button).

src/south/FolderScanner/FolderScanner.class.js
```diff
@@ -25,6 +25,10 @@ class FolderScanner extends ProtocolHandler {
     this.minAge = minAge
     this.regex = new RegExp(regex)
     this.compression = compression
+
+    this.lastOnScanAt = null
+    this.lastAddFileAt = null
```
jfhenon, line 30: this.lastOnScanAt = null
    this.lastAddFileAt = null
could be part of the ProtocolHandlerClass

src/south/FolderScanner/FolderScanner.class.js
```diff
@@ -34,6 +38,7 @@ class FolderScanner extends ProtocolHandler {
    */
   async onScan(scanMode) {
     this.logger.debug(`FolderScanner activated on scanMode: ${scanMode}.`)
+    this.lastOnScanAt = new Date().getTime()
```
jfhenon, line 41: This variable can be maintained by the parent class ProtocolHandler (before calling onScan)

src/south/FolderScanner/FolderScanner.class.js
```diff
@@ -126,6 +131,18 @@ class FolderScanner extends ProtocolHandler {
       await this.setConfig(filename, `${stats.mtimeMs}`)
     }
   }
+
+  /**
```
jfhenon, line 135: Nothing special for this class. The default method could be used.

src/south/ProtocolHandler.class.js
```diff
@@ -174,6 +176,16 @@ class ProtocolHandler {
       setTimeout(resolve, timeout)
     })
   }
+
+  /**
+   * Get live status.
+   * @returns {object} - The live status
+   */
+  getStatus() {
```
jfhenon, line 184: The default implementation should not be surcharged and could return standard variables. 

if we need a specific variable for this handler:

getStatus() {
const status = super.getStatus()
status.anykey = anyvalue
return status
}

jfhenon:
@attilasuto 
image
I would either put the (i) close to the name UA or I would turn it into a normal button close to the points button.
@burgerni10 your view?

src/south/ProtocolHandler.class.js
```diff
@@ -174,6 +180,19 @@ class ProtocolHandler {
...
+    const status = { 'Last scan time': this.lastOnScanAt ? new Date(this.lastOnScanAt).toLocaleString() : 'Never' }
+    if (this.handlesFiles) {
+      status['Last file added time'] = this.lastAddFileAt ? new Date(this.lastAddFileAt).toLocaleString() : 'Never'
+      status['Number of files added'] = this.addFileCount
+    }
```
jfhenon, line 193: need to also track Last value and number of values, last timestamp, last pointid,

jfhenon (changes requested):
Made some comment on UI 
Also on default status for values

burgerni10:
> @attilasuto
> image
> I would either put the (i) close to the name UA or I would turn it into a normal button close to the points button.
> @burgerni10 your view?

I am not sure about this.

The way it is now, it will always display the (i) at the same place, making UI more consistent. Next to the name, it would change a little bit regarding the length of the breadcrumb.

But I agree with the second part of your message : it should look like a normal button. The way it is designed now make it look like the help button (?) or something where a popover appears once we click on it. @attilasuto, what do you think ? 

I suggest a button with "Status" in it.

src/south/ProtocolHandler.class.js
kukukk: I was planning to include point related informations when we implement it for the first point compatible South.

attilasuto:
> > @attilasuto
> > image
> > I would either put the (i) close to the name UA or I would turn it into a normal button close to the points button.
> > @burgerni10 your view?
> 
> I am not sure about this.
> 
> The way it is now, it will always display the (i) at the same place, making UI more consistent. Next to the name, it would change a little bit regarding the length of the breadcrumb.
> 
> But I agree with the second part of your message : it should look like a normal button. The way it is designed now make it look like the help button (?) or something where a popover appears once we click on it. @attilasuto, what do you think ?
> 
> I suggest a button with "Status" in it.

I agree it would fit more better to have a button with "Status" on the settings screen and keep the (i) on the south listing screen. I will change this as required.

src/south/ProtocolHandler.class.js
jfhenon: I think the default status could be added now as well.

jfhenon (commented):
@kukukk minor comments on ProtocolHandler
line 44 -> this.compressionLevel should this be an engine parameter?
line 66 -> maybe we could avoid to ask the south to use super.onScan by asking the South handler to define onScanImplementation that would be call by OnScan.

jfhenon (approved):

burgerni10:
I approve the Pr, but I suggest some changes in issue #1173  
I put these changes in a separate issue because it's more general than just the "Status" button

burgerni10 (approved):

PR #1174, opened 2021-03-24
Modbus/reconnection

burgerni10:


burgerni10:
Hi @Lulu300, can you check if this PR fix the reconnection problem with Modbus ?

Lulu300:
Hi, I've just tried and I can confirm that's working fine! Thank you for implementing this feature! I think there is just a small issue in the migration rule I'll annotate what I think is an issue

src/migration/migrationRules.js
```diff
@@ -393,18 +365,10 @@ module.exports = {
...
-      if (dataSource.protocol === 'Modbus') {
-        dataSource.points.forEach((point) => {
-          if (!Object.prototype.hasOwnProperty.call(point, 'multiplierCoefficient')) {
-            logger.info('Add multiplierCoefficient field to Modbus points')
-            point.multiplierCoefficient = 1
-          }
-        })
+        if (!Object.prototype.hasOwnProperty.call(dataSource.Modbus, 'retryInterval')) {
+          logger.info('Add retryInterval field to Modbus')
+          dataSource.Modbus.swapWordsInDWords = 10000
```
Lulu300, line 370: I think it should be dataSource.Modbus.retryInterval instead of dataSource.Modbus.swapWordsInDWord

PR #1176, opened 2021-03-26
fix(south-sqldbtofile): manage lastCompletedDate from string date result

burgerni10:


src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -93,6 +93,26 @@ class SQLDbToFile extends ProtocolHandler {
     }
   }
 
+  /**
+   * Function used to parse an entry and update the lastCompletedAt if needed
+   * @param {*} entry - on sql result item
+   * @return {void}
+   */
+  checkLastCompletedAt(entry) {
+    if (entry[this.timeColumn] && (entry[this.timeColumn] instanceof Date) && (entry[this.timeColumn] > new Date(this.lastCompletedAt))) {
```
jfhenon, line 102: isn't it enought to  (entry[this.timeColumn] instanceof Date

src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -93,6 +93,26 @@ class SQLDbToFile extends ProtocolHandler {
...
+      this.lastCompletedAt = entry[this.timeColumn].toISOString()
+    } else if (entry[this.timeColumn]) {
```
jfhenon, line 104: If I read correctly, nothing happens if entry[this.timeColumn] is undefined or null should we do a silly log?

src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -93,6 +93,26 @@ class SQLDbToFile extends ProtocolHandler {
...
+      const entryDate = new Date(entry[this.timeColumn])
+      if (entryDate.toString() !== 'Invalid Date') {
```
jfhenon, line 106: same question: if we receive Invalid Date, should we silly log?

src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -93,6 +93,26 @@ class SQLDbToFile extends ProtocolHandler {
...
+        // We need to take back the js added timezone since it is not in the original string coming from the database
+        const entryDateWithouthTimezonOffset = new Date(entryDate.getTime() - entryDate.getTimezoneOffset() * 60000)
```
jfhenon, line 108: Without not Withouth

jfhenon: pour passer en UTC, il n'y a pas la fonction Date.UTC()?

src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -134,9 +154,7 @@ class SQLDbToFile extends ProtocolHandler {
 
     if (result.length > 0) {
       result.forEach((entry) => {
-        if (entry[this.timeColumn] && (entry[this.timeColumn] instanceof Date) && (entry[this.timeColumn] > new Date(this.lastCompletedAt))) {
-          this.lastCompletedAt = entry[this.timeColumn].toISOString()
-        }
+        this.checkLastCompletedAt(entry)
```
jfhenon, line 157: I would call it set LastCompletedAt and maybe it would be clearer to wrtie:
this.lastCompletedAt = this.setLastCompletedAt(entry)

src/south/SQLDbToFile/SQLDbToFile.class.js
burgerni10: I don't think we should: this function is called for every entry of a query results. If it's a big query, the logger may overload the logs. Even if it's a silly mode, I'd prefer not to.

src/south/SQLDbToFile/SQLDbToFile.class.js
burgerni10: Date.UTC takes in argument several number (year, month, day, etc... : https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC)
Numbers that I don't have without parsing the string date into a date format which is what I do line 105. And the Date constructor give an offset depending of the runtime environment timezone. That is why I remove the offset in the constant.

Issue #1140, opened 2021-03-08
MQTT - fail to listen to a point when it's listed twice


burgerni10:
When a point is listed twice in the MQTT point list, we receive no value at all.

The first subscription should be done, and the second one is ignored, which triggers a warning in the logger.

burgerni10:
We need to manage when the connector subscribes to a point with a wildcard AND a specific subscription

kukukk:
How should it be handled? Possible options I see:
  - check on the UI if there are overlapping points and disable saving until the user fixes it
  - simply ignore the second point and skip the subscription
  - identify and subscribe only to the higher impact point

Simply ignoring the second one could cause data loss. E.g.: '/sensors/temperature1' vs '/sensors/*'. The second one could include tens of sensor values, but we will ignore it and only receive data from a single sensor.

burgerni10:
Your first point must indeed be done. This logic may be used again in other connector (like OPC) in a later issue.

I think the third point solve the issue without data loss, so you can implement that and forget about the second point.

burgerni10:
The backend part will be done with the issue #1104 

@attilasuto can you implement the frontend please ?

attilasuto:
Please check merge request: #1247

The points config page will show error for any duplicated entry that has the 'unique = true' defined in the schema. Currently I've added the unique flag to pointId on MQTT.schema. This can be easily enabled on any protocol that requires a unique flag.
Please let me know if any change is required.

PR #1247, opened 2021-06-03
feat(points): Error on MQTT duplicated points.

attilasuto:
For task: #1140

duplicate
error
schema


burgerni10 (changes requested):
[Edited] : issue #1251 created regarding this comment. @attilasuto , can you instead improve the MQTT help to explain that it's the user's responsibility to check there is no duplicate if he uses wildcards (# or +) ?

Your PR works fine on identical points, but this tricky use case is still a problem : 
image

The last point covers the first one and it is seen as a duplicate. 
Can you please improve the validation method for MQTT by checking this case too ?

attilasuto:
Help added, the improved validation will be done separately for task #1251

burgerni10 (approved):

Issue #1055, opened 2021-01-21
OPCUA scan group name on OPCUA point list


burgerni10:
OPCUA scan group unicity must be made with the "scanMode - aggregate - resampling (if any)" and this should appears on the scan mode for OPCUA point lists

burgerni10:
Keep only the scan modes set in scan groups

burgerni10:
Following some internal discussion : 
-This issue only applies for Scan Group protocols (OPCUA_HA, OPCHDA)
-In such protocols, a scan mode without scan groups (i.e. aggregate and/or resampling) cannot be used

To sum it up : in OPCUA_HA and OPCHDA, when listing points, only scan modes set in a scan group can be displayed.

To guide the user for these protocols, a help can be added to tell the user that only scan modes set in scan groups are displayed ; a new scan group must be added to be able to select other scan modes.

PR #1250, opened 2021-06-07
Scan group protocols should only allow scan modes set in a scan group as option

attilasuto:
#1055

### DIFF
--- a/src/client/South/ConfigurePoints.jsx
+++ b/src/client/South/ConfigurePoints.jsx
@@ -171,7 +171,7 @@ const ConfigurePoints = () => {
     const name = `points.${index}.${key}`
     return (
       /* eslint-disable-next-line react/jsx-props-no-spreading */
-      { name, value: <Control onChange={onChange} name={name} point {...rest} /> }
+      { name, value: <Control onChange={onChange} name={name} {...rest} /> }
     )
   }))
 
--- a/src/client/South/ConfigurePoints.jsx
+++ b/src/client/South/ConfigurePoints.jsx
@@ -171,7 +171,7 @@ const ConfigurePoints = () => {
     const name = `points.${index}.${key}`
     return (
       /* eslint-disable-next-line react/jsx-props-no-spreading */
-      { name, value: <Control onChange={onChange} name={name} {...rest} /> }
+      { name, value: <Control onChange={onChange} name={name} point {...rest} /> }
     )
   }))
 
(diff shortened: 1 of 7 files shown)


src/client/South/ConfigurePoints.jsx
```diff
@@ -171,7 +171,7 @@ const ConfigurePoints = () => {
     const name = `points.${index}.${key}`
     return (
       /* eslint-disable-next-line react/jsx-props-no-spreading */
-      { name, value: <Control onChange={onChange} name={name} {...rest} /> }
+      { name, value: <Control onChange={onChange} name={name} point {...rest} /> }
```
burgerni10, line 174: I'm not sure to understand the 'point' props : in OIbScanMode component, it is marked as a boolean but here it seems to me it is an object.

src/client/components/OIbForm/OIbScanMode.jsx
```diff
@@ -1,12 +1,19 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import { useParams } from 'react-router-dom'
 import { FormGroup, FormText, Label, Input, FormFeedback } from 'reactstrap'
 import { ConfigContext } from '../../context/configContext.jsx'
 
-const OIbScanMode = ({ label, help, valid, value, name, onChange }) => {
+const OIbScanMode = ({ label, help, valid, value, name, onChange, point }) => {
```
burgerni10, line 7: See my previous comment on the point type (file ConfigurePoint)

src/client/components/OIbForm/OIbScanMode.jsx
```diff
@@ -1,12 +1,19 @@
...
   const { newConfig } = React.useContext(ConfigContext)
   const scanModes = newConfig?.engine?.scanModes ?? [] // scan modes defined in engine
-  const options = scanModes.map((e) => e.scanMode)
+  const { dataSourceId } = useParams()
+  const dataSource = newConfig?.south.dataSources.find((d) => d.dataSourceId === dataSourceId)
+  const scanGroups = dataSource ? dataSource[dataSource.protocol]?.scanGroups : null
+  // Scan Group protocols should only allow scan modes set in a scan group as option
+  const options = scanGroups !== undefined && point
```
burgerni10, line 14: My guess is that you want to display scanGroups only in point list window. Maybe this props should have another name to better understand its role.
Please enlighten me if I misunderstand this props

burgerni10 (changes requested):

src/client/components/OIbForm/OIbScanMode.jsx
attilasuto: Yes, exactly this is only needed when configuring a new point. Maybe we can use another naming like: isPointConfig.

attilasuto: I have removed the point flag.

burgerni10 (approved):

PR #1268, opened 2021-06-24
feat(bulk): Refactor Listen type South

kukukk:


src/south/ListenProtocolHandler.class.js
```diff
@@ -0,0 +1,51 @@
...
+ * driver will be able to "listen" for updated values.
+ * - **connect**: to allow to establish proper connection to the data source(optional)
+ * - **disconnect**: to allow proper disconnection (optional)
+ * In addition, it is possible to use a number of helper functions:
+ *
+ * All other operations (cache, store&forward, communication to North applications) will be
+ * handled by the OIBus engine and should not be taken care at the South level.
+ *
+ */
+class ListenProtocolHandler extends ProtocolHandler {
```
jfhenon, line 17: I thought we would not create another parent class. I think it may limit the south developers. For example,  how do create a south that can support both listening to points and reading to the last point? (such as OPCUA)

kukukk: We can keep one single parent class, but it will be difficult to know which methods should be overwritten when implementing a new South.
For OPCUA we have separate implementations for OPCUA_DA and OPCUA_HA. If we want to implement a LastPoint OPCUA we could create a third OPCUA South.
Mixing different functionalities makes harder to understand and maintain the code.

jfhenon: on the other hand, every "base" improvement we do in UA (for example, new method of identification), have to be replicated in mulitple places.

burgerni10: I agree that OPCUA_DA should have both (lastPoint as it is for now, and listen that will be developed later). In this case, JF suggests to check in the parent class if the function is implemented.

The constructor could be something like : 
class MQTT extends ProtocolHandler {
constructor ()
{
super({supportListen: true, supportLastPoint: false, supportHistory: false})

And the check in the parent class could be like that : 
if (listenSupport) if (typeof listen !== “function”) ... log erreur

src/south/ListenProtocolHandler.class.js
```diff
@@ -0,0 +1,51 @@
...
+   * Constructor for ListenProtocolHandlerClass
+   * @constructor
+   * @param {*} dataSource - The data source
+   * @param {Engine} engine - The engine
+   * @return {void}
+   */
+  constructor(dataSource, engine) {
+    super(dataSource, engine)
+
+    this.handlesPoints = true
```
burgerni10, line 28: Following my answer to JF's comment, this must be in the ProtocolHandler class

src/south/ListenProtocolHandler.class.js
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * The listen method implements the subscription for the points.
+   * @param {object} _data - The data required to configure listening
+   * @returns {void}
+   */
+  listen(_data) {
+    const { dataSourceId } = this.dataSource
+    this.logger.error(`Data source ${dataSourceId} should surcharge listen()`)
+  }
```
burgerni10, line 48: This won't be needed anymore if the check is done as mentionned in my answer to JF's comment. Therefore you ca nremove this class

burgerni10 (changes requested):

src/south/ListenProtocolHandler.class.js
kukukk: Since we define listen() in ProtocolHandler (logging that it should be overwritten by the South implementation) it will always be a function and this check doesn't help us.

jfhenon: unless we remove the default listen() from ProtocolHandler.

kukukk: So do you want to remove those the listen(), lastPointQuery() and historyQuery() methods from ProtocolHandler and just document it somewhere that a new South implementation must implement some of them (based on the what type of South it is), and this way we could add the check if the required method is implemented?
I'm not sure yet, if this.listen is already defined when calling the parent constructor, but I will check it.

PR #1270, opened 2021-06-24
fix-s3

marouanehassanioptimistik:


marouanehassanioptimistik:
@burgerni10 i have copied the method that was moved and unaccessible from the s3 class. It is not very dry though we can probably do better by sharing utils function with a util class or exposing this method elsewhere. I let you decide if where this function should be put or if we should leave it like that.

burgerni10:
> @burgerni10 i have copied the method that was moved and unaccessible from the s3 class. It is not very dry though we can probably do better by sharing utils function with a util class or exposing this method elsewhere. I let you decide if where this function should be put or if we should leave it like that.

Some refactoring will be done in future release on how the connectors interact with the Engine and other services. I let this function there for now but I note that it will be move later into a util class

PR #1281, opened 2021-07-05
feat(bulk): Refactor LastPoint type South

kukukk:


src/south/FolderScanner/FolderScanner.class.js
```diff
@@ -15,7 +15,7 @@ class FolderScanner extends ProtocolHandler {
    * @return {void}
    */
   constructor(dataSource, engine) {
-    super(dataSource, engine)
+    super(dataSource, engine, { supportListen: false, supportLastPoint: true, supportHistory: false })
```
jfhenon, line 18: should we have a supportLastFile? or supportFile?

jfhenon (commented):
see comment for File support mode

src/south/FolderScanner/FolderScanner.class.js
kukukk: We have **this.handlesPoints** and **this.handlesFiles** set in every South constructor (used in **getStatus())**. We could use those variables to identify whether the given South sends data as values or file and rename **lastPointQuery()** to **lastFileQuery()** for FolderScanner.

burgerni10: What about changing supportLastPoint to supportScanMode ? It's more straightforward and point/file abstract.

lastPointQuery would be renamed scanModeQuery.

About the point/file dichotomy, I think we can eventually merge them in the South connectors. I propose we keep it like that for now, I need to think a bit more about it.

burgerni10 (approved):

Issue #1256, opened 2021-06-11
In home page, North/South names can overlap


brunomaury:
![image](https://user-images.githubusercontent.com/56166954/121687539-c1462d00-cac2-11eb-9b22-543539f1c99f.png)

PR #1290, opened 2021-07-07
fix(ui): fix overllaping text in home screen #1256

franckfotso1:
oibus05


burgerni10 (approved):

burgerni10:
@jfhenon do you agree with the ellipsis method ?
I've discussed it with Bruno, it may be interesting to have an overlay on mouse hover.

PR #1296, opened 2021-07-08
fix(ui): warning-user-default-password #1275

franckfotso1:
oibus_changing01
oibus_changing
oibus_password_warning


burgerni10 (changes requested):
change password -> Change password

burgerni10 (approved):

Issue #1298, opened 2021-07-08
[UI] - Add overlay for node title on mouse hover


burgerni10:
Add an overlay when the mouse is over a node connector to display the full name of the connector

![image](https://user-images.githubusercontent.com/28055217/124960308-77e2f200-e01c-11eb-9d30-1fc926e4dfd1.png)

PR #1301, opened 2021-07-09
fix(ui): mouse-hover-overlay #1298

franckfotso1:


src/client/Health/NodeView.jsx
```diff
@@ -44,7 +44,7 @@ const NodeView = ({ status, onRestart, onShutdown }) => {
         label: (
           <div className={`oi-box tight text-${application.enabled ? 'success' : 'muted'}`}>
             <Link to={`/north/${application.applicationId}`}>
-              <div style={{ overflow: 'hidden', textOverflow: 'ellipsis' }}>{application.applicationId}</div>
+              <div id="overlay">{application.applicationId}</div>
```
burgerni10, line 47: Why do you use an id ? Id should be use uniquely, but here you will have several id in your code with the same reference. What you need is a class, not an id

src/client/Health/NodeView.jsx
```diff
@@ -79,7 +79,7 @@ const NodeView = ({ status, onRestart, onShutdown }) => {
         label: (
           <div className={`oi-box tight text-${dataSource.enabled ? 'success' : 'muted'}`}>
             <Link to={`/south/${dataSource.dataSourceId}`}>
-              <div style={{ overflow: 'hidden', textOverflow: 'ellipsis' }}>{dataSource.dataSourceId}</div>
+              <div id="overlay">{dataSource.dataSourceId}</div>
```
burgerni10, line 82: Why do you use an id ? Id should be use uniquely, but here you will have several id in your code with the same reference. What you need is a class, not an id

src/client/style/oi.css
```diff
@@ -429,3 +429,12 @@ h3 {
 .points .table td .oi-icon {
   margin-top: 9px;
 }
+
+#overlay:hover {
```
burgerni10, line 433: Since it's a custom class, describe it .overlay and not #overlay (which is used for id)

Moreover, to better identify custom classes, you must write it oi-overlay (oi being the shortcut for optimistik)

src/client/style/oi.css
```diff
@@ -429,3 +429,12 @@ h3 {
...
+  overflow: visible;
+	}
+
+#overlay {
```
burgerni10, line 437: Don't forget to update this name too

burgerni10 (changes requested):

burgerni10:
After a test on my machine, it doesn't seem to fix the issue : 
image

I suggest to forget this PR for now, and we can focus on the draft PR #1302 later as described in its comments

Issue #1191, opened 2021-03-30
When renaming a South/North, the corresponding sqlfile table is not renamed which potentially can cause loss of data.


jfhenon:
either
1/ we make sure the cache is empty before allowing a rename
2/ we rename the file as well (preferred I think)

kukukk:
Hmm... Not sure if it so easy. The rename is done in the frontend. The backend knows nothing about the config change, just reads the actual config after restart. 

Maybe it would be better to use an additional `id`, which is hidden from the user. It should be generated by the frontend when a new North/South is added using UUID - which can be considered to be unique. This `id` would be the same after the North/South is renamed. The backend could use this `id` to identify the North/South in tables and to generate SQLite filenames.
We could also write a migration, which generates this `id` for every North/South and makes the necessary renames.

jfhenon:
The frontend use the name in the URL as well so I don’t see this hidden Id as a preferred solution. Is it possible to ask for a rename to the backend? 
JF

Jean-François HÉNON - Optimistik
+33 6 03 79 73 26

kukukk:
We could have a separate `id` and `name`, where `id` would be used by the backend and `name` by the frontend. We could even keep using applicationId and dataSourceId for `name`, so the frontend changes would be minimal (just generate the `id` when creating a new North/South)

Asking the backend for a rename could have some extra complications:
 - if we do this during runtime operation, the Cache must close the SQLite connection so the file can be renamed, but what do we do with new values during this time? 
 - if we do this with a restart of OIBus, we have to store somewhere the information about which North/South should be renamed, and handle it before the worker starts (similar to migrations).

Also, using a simple free text input as a source for the filename may not be a very good idea, because the user can introduce any special characters. Having a UUID `id` would eliminate this possible problem.

burgerni10:
I totally agree with the `uuid` solution. 

After some talks with JF, his main concerns are about the simplicity of use in general (usage and debug). 
Being able to see to which connector each sqlite database is linked is useful. I suggest that the uuid can be displayed in the connector (and maybe in the south / north connector list) so if needed we can look into the sqlite database (can be done in another issue)
For the url, I don't think it is a problem to have an uuid instead of the name of the connector since it will still be displayed on the page thanks to the breadcrumb. Speaking of which, the breadcrumb should be reworked a bit to take this change into account (using the uuid in url and name for display)

[...]

PR #1302, opened 2021-07-09
feature(nodeview): add tooltip

burgerni10:


burgerni10:
This is a draft to fix the issue #1298 

For this to be finalised, the dataSourceId must be unique and css compatible (no space for example). I suggest we wait for the issue #1191 to be done before going back to this one.

jfhenon (changes requested):
check the use of random in South/North loops.

jfhenon (approved):

burgerni10:
I moved the targeted id of each tooltip to the main div instead to have it in the div having the applicationId/dataSourceId : it is more reactive this way when the mouse comes hover the node.

There was a small error in the css with the shadow.

Lastly, I rebased the branch and commits to have a clean working tree.

@franckfotso1 check the corrections when you will have time

PR #1304, opened 2021-07-12
Live info status

franckfotso1:


src/north/oianalytics/OIAnalytics.class.js
```diff
@@ -60,6 +61,8 @@ class OIAnalytics extends ApiHandler {
   async handleFile(filePath) {
     const stats = fs.statSync(filePath)
     this.logger.silly(`OIAnalytics handleFile(${filePath}) (${stats.size} bytes)`)
+    this.statusData['Last uploaded file'] = filePath
```
burgerni10, line 64: Can you add another field for the date ? Like 'Last file upload at' ?

src/north/ApiHandler.class.js
```diff
@@ -55,6 +59,26 @@ class ApiHandler {
   connect() {
     const { name, api } = this.application
     this.logger.info(`North API ${name} started with protocol ${api}`)
+    this.statusData['Connected at'] = new Date().toISOString()
+    this.updateStatusDataStream()
```
burgerni10, line 63: I disagree with this : some connectors do not 'connect' like fileWriter for example. Moreover, when a connector connects to something, you should not update this status here : the connection occurs generally on the driver (like for protocols) so you should update the statusData in the callback function instead for each api (when applicable)

src/north/ApiHandler.class.js
```diff
@@ -65,6 +89,18 @@ class ApiHandler {
   disconnect() {
     const { name, id } = this.application
     this.logger.info(`North API ${name} (${id}) disconnected`)
+    this.engine.eventEmitters[`/north/${id}/sse`]?.events?.off('data', this.listener)
+    this.statusData['Connected at'] = 'Not connected'
```
burgerni10, line 93: Same remark than for the 'connect' method

src/north/ApiHandler.class.js
```diff
@@ -45,6 +47,8 @@ class ApiHandler {
     const { logParameters } = this.application
     this.logger = new Logger()
     this.logger.changeParameters(this.engineConfig, logParameters)
+    this.statusData = {}
+    this.sseData = {}
```
burgerni10, line 51: Why do you need sse ?

src/north/ApiHandler.class.js
```diff
@@ -65,6 +89,18 @@ class ApiHandler {
...
     const { name, id } = this.application
     this.logger.info(`North API ${name} (${id}) disconnected`)
+    this.engine.eventEmitters[`/north/${id}/sse`]?.events?.off('data', this.listener)
+    this.statusData['Connected at'] = 'Not connected'
+    this.updateStatusDataStream()
+  }
+
+  listener = (data) => {
+    if (data) this.sseData = data
+    this.engine.eventEmitters[`/north/${this.application.id}/sse`]?.stream?.write(`data: ${JSON.stringify(this.sseData)}\n\n`)
```
burgerni10, line 99: Why is it different than protocolHandler ? You don't need sse and you should check if data exists (see listener method in ProtocolHandler.class.js)

src/north/console/Console.class.js
```diff
@@ -27,8 +27,12 @@ class Console extends ApiHandler {
   async handleValues(values) {
     if (this.verbose) {
       console.table(values, ['pointId', 'timestamp', 'data'])
+      this.statusData['Last handled Values at'] = new Date().toISOString()
+      this.updateStatusDataStream()
```
burgerni10, line 31: You can simplify this by updating the status outside the if block (and remove the call in the else block.

Moreover, maybe you can display the last value of the array (content and pointId) like it is done in the protocol connectors

src/north/console/Console.class.js
```diff
@@ -47,6 +51,8 @@ class Console extends ApiHandler {
       fileSize,
     }]
     console.table(data)
+    this.statusData['Last uploaded file'] = filePath
```
burgerni10, line 54: Can you add a 'Last upload at' to display the date ? Please do so everywhere you have the same logic

src/north/console/Console.class.js
burgerni10: Please do so in every north connectors

src/south/OPCUA_DA/OPCUA_DA.class.spec.js
```diff
@@ -21,6 +21,8 @@ jest.mock('../../engine/Logger.class')
 // Mock engine
 const engine = jest.createMockFromModule('../../engine/Engine.class')
 engine.configService = { getConfig: () => ({ engineConfig: config.engine }) }
+engine.eventEmitters = {}
+engine.eventEmitters = {}
```
burgerni10, line 25: Remove one of the declaration

burgerni10 (changes requested):

burgerni10 (changes requested):
After some tests, here are some improvements to make : 

- [x] When a connector is not enabled, an error is displayed. Can you fix that by checking on the status component initialization if it's enabled (in this case you subscribe to the socket) or not (in this case a message should appear to tell the user why no information will be displayed)
- [x] It seems that the first display of status arrives when a new event is emitted from the backend. It can be really long if the scan mode is every10Hours for example. Maybe an event should be emitted on socket subscription so we do not wait for the next event to occur
- [x] On the frontend, the dates are displayed in iso string. It may be nice to have them in local format. 
- [x] On the Home page, you can add other status information (inside the engine) by displaying several columns and use the sse system to retrieve the valeus automatically

burgerni10 (approved):

Issue #1154, opened 2021-03-13
Add health data to the log every x minutes (maybe every 120 minutes)


jfhenon:
It's nice to see health information on the main screen but when you debug for something that happened a long time ago, it's difficult to get this information.

PR #1306, opened 2021-07-12
feat(logs): add loki logs

burgerni10:


burgerni10:
Working poc to send data to loki : 

image

For now, I take the engineName, so if we set up loki logs, we need to rename the oibus engine so we can differentiate them.

To do : 
-basic auth
-send the logs to another OIBus
-get the logs from another OIBus and send them to loki (or to another OIBus ?)

src/engine/Logger.class.js
```diff
@@ -31,16 +32,44 @@ class Logger {
...
+        fileName: sqliteLog.fileName,
+        level: sqliteLog.level,
+        maxFileSize: sqliteLog.maxSize,
+        handleExceptions: true,
+      }))
+    }
+
+    if (lokiLog.level !== 'none') {
+      this.logger.add(new LokiTransport({
+        host: lokiLog.host, // http://loki.igw.rd.optimistik.fr:3100
```
marouanehassanioptimistik, line 65: please remove this as it will spread our loki url

marouanehassanioptimistik (changes requested):

src/south/ProtocolHandler.class.js
```diff
@@ -47,7 +47,7 @@ class ProtocolHandler {
 
     const { logParameters } = this.dataSource
     this.logger = new Logger()
-    this.logger.changeParameters(this.engineConfig.logParameters, logParameters)
+    this.logger.changeParameters(this.engineConfig.engineName, this.engineConfig.logParameters, logParameters)
```
jfhenon, line 50: maybe we should only send the engineConfig and it's up to the called function to pick engineName and logParameters.

jfhenon (commented):
- Should the unique ID for Loki be a parameter at the Loki config level (making the OIBus engine name unique among all users may not be optimal).
- We don't solve the case of OIBus with no access to Internet. We should add the ability to forward Logs to another OIBus
- We may not need the OIBus alive anymore (or limit it to a regular log)

burgerni10:
> * Should the unique ID for Loki be a parameter at the Loki config level (making the OIBus engine name unique among all users may not be optimal).
=> I don't see the point of the engineName anyway : should we juste have a 'loki-id' in the loki log level and remove the engine name from the config ?

> * We don't solve the case of OIBus with no access to Internet. We should add the ability to forward Logs to another OIBus
=> Will be done in a following PR

> * We may not need the OIBus alive anymore (or limit it to a regular log)
=> I agree that the Alive signal is OIAnalytics specific : people not using OIA must implement an endpoint for it. So for me it makes sense to remove the alive signal, once issue #1154 is done (it makes sense to do it now that winston-loki is almost implemented)

jfhenon (approved):

marouanehassanioptimistik (approved):

PR #1309, opened 2021-07-16
feat(health-signal): Health status send to logs

burgerni10:


burgerni10:
PR based on #1306

src/client/Engine/Engine.validation.js
```diff
@@ -38,11 +38,17 @@ const validation = {
...
-    aliveSignal: {
-      host: notEmpty('Host'),
-      endpoint: notEmpty('Endpoint'),
-      id: notEmpty('Id'),
-      frequency: inRange(60, 3600),
+    healthSignal: {
+      http: {
+        host: notEmpty('Host'),
+        endpoint: notEmpty('Endpoint'),
+        id: notEmpty('Id'),
```
jfhenon, line 45: An engine will have 3 identifications:
- engine name
- http id
-  logging id

What is needed for this id is to have unicity so we can recognise the OIBus at the origin of the signal/log. We can't easily impose the unicity of engine names.

I propose 2 solutions: 
1 -> generate a unique id for each installation (could be based on the timestamp of the installation, or just a uuid). This will create unicity but it will be necessary to identify on the server what is the OIBus corresponding to a given ID. 

2-> Create the concept of a "domain" (could be named federation, network name, customer name, etc). The unique id would be domain-engine name. For example "optimistik-oibus"

src/engine/Engine.class.js
```diff
@@ -97,7 +97,7 @@ class Engine {
     this.addValuesMessages = 0
     this.addValuesCount = 0
     this.addFileCount = 0
-    this.forwardedAliveSignalMessages = 0
+    this.forwardedHealthSignalMessages = 0
```
jfhenon, line 100: this attributes should be managed by the object HealthSignal.

jfhenon (commented):
need to think about ids

src/client/Engine/Engine.validation.js
burgerni10: I think we can use the engineName only: 

-When updating OIBus to the future version with the improved logs, we must tell the client to update the engineName to something meaningful: 'OIBus' which is the default name to 'Optimistik demo' for example

-When installing a new OIBus: we can introduce a step in the installation process to give the engine name. I propose that we do this installation update when we'll improve the whole installation process. In the meantime we do the same  than for an update

src/engine/Engine.class.js
burgerni10: This field is not used anyway. I remove it for now and I will implement it with the forward feature that I will do in another PR

burgerni10: Will be done with issue #1325

Issue #1326, opened 2021-07-21
[Archive mode] - Allow the user to set a retention duration when archive mode is activated


burgerni10:
The retention period must be a number of hours. 
When migrating the config, if archive mode is selected, used a default retention period of one year (8760 hours)

Moreover, it may be better to hide the retention duration AND the archive folder path when the archive mode is deactivated.

Instead of having an archive vs delete mode, replace this select by a boolean

On the backend, a job must run once a day to empty the archive folder if the files are older than the retention period

burgerni10:
![image](https://user-images.githubusercontent.com/28055217/126504235-aeb08f80-a9ed-4a2e-90ef-68bfc0548e32.png)

PR #1334, opened 2021-07-28
feat(ui): add retention duration for archive mode #1326

franckfotso1:
Capture d’écran 2021-07-27 à 16 38 23
Capture d’écran 2021-07-27 à 16 38 50


src/client/Engine/Caching.jsx
```diff
@@ -35,30 +35,45 @@ const Caching = ({ onChange, caching }) => (
...
-          value={caching.archiveMode}
-          defaultValue="archive"
-          help={<div> Move to Archive Folders or Delete files</div>}
+        <OIbInteger
+          name="engine.caching.archive.retentionDuration"
+          label="Retention Duration"
+          value={caching.archive.retentionDuration}
+          defaultValue={8760}
+          valid={validation.engine.caching.archive.retentionDuration}
+          help={<div>How long to wait for the files to get delete in the archive folder (in hours)</div>}
```
burgerni10, line 71: How long to wait for the files to get delete in the archive folder (in hours) -> Retention period of archived files (in hours)

src/config/defaultConfig.json
```diff
@@ -18,9 +18,12 @@
       "sqliteMaxFileSize": 50000000
     },
     "caching": {
-      "cacheFolder": "./cache",
-      "archiveMode": "archive",
-      "archiveFolder": "./cache/archived/"
+      "archive": {
+        "enabled": true,
+        "archiveFolder": "./cache/archived/",
```
burgerni10, line 23: Replace 'archived' by archive

src/engine/Cache.class.js
```diff
@@ -421,36 +428,78 @@ class Cache {
...
-          break
-        default:
-          this.logger.error(`unknown Archive Mode: ${this.archiveMode}`)
+  /**
+   * Delete file in archiveFolder
+   * @param {string} filePath - The file
+   * @return {void}
+   */
+  async refreshArchiveFolder() {
+    this.logger.silly('refreshArchiveFolder()')
```
burgerni10, line 464: This log is a little bit dirty. Can you rephrase it ? Something like 'Looking for archived files to remove'
And change it from silly to debug

src/engine/Cache.class.js
```diff
@@ -421,36 +428,78 @@ class Cache {
...
+    const timestamp = new Date().getTime()
+    if (files.length > 0) {
+      // Disable ESLint check because we need for..of loop to support async calls
+      // eslint-disable-next-line no-restricted-syntax
+      for (const file of files) {
+        // Disable ESLint check because we want to handle files one by one
+        // eslint-disable-next-line no-await-in-loop
+        const stats = fs.statSync(path.join(this.archiveFolder, file))
+
+        if (stats.mtimeMs + this.retentionDuration < timestamp) {
```
burgerni10, line 480: this.retentionDuration is in seconds (you multiply it by 3600 in the constructor).  stats.mtimeMs give you milliseconds. You must change multiply by 3600000 in the constructor instead

src/engine/Cache.class.js
```diff
@@ -421,36 +428,78 @@ class Cache {
...
+            this.logger.error(`Error sending the file ${file}: ${sendFileError.message}`)
+          }
+        }
       }
+    } else {
+      this.logger.debug(`The archive folder ${this.archiveFolder} is empty. Nothing to delete`)
     }
+    this.archiveTimeout = setTimeout(() => {
+      this.refreshArchiveFolder()
+    }, 3600000)
```
burgerni10, line 502: Create a global const variable at the top of the file and use it here. It's easier if we need to change the duration later or just if we want to check its value without looking for the timeout

src/client/Engine/Caching.jsx
```diff
@@ -35,30 +35,45 @@ const Caching = ({ onChange, caching }) => (
...
-          name="engine.caching.archiveMode"
-          options={['archive', 'delete']}
-          value={caching.archiveMode}
-          defaultValue="archive"
-          help={<div> Move to Archive Folders or Delete files</div>}
+        <OIbInteger
+          name="engine.caching.archive.retentionDuration"
+          label="Retention Duration"
+          value={caching.archive.retentionDuration}
+          defaultValue={8760}
```
burgerni10, line 69: Thinking back at this value, I think one year is too big. Can you make it to one month (30 days) instead ?

src/migration/migrationRules.js
```diff
@@ -486,4 +486,13 @@ module.exports = {
       }
     })
   },
+  25: (config) => {
+    config.engine.caching.archive = {
+      enabled: config.engine.caching.archiveMode === true,
+      archiveFolder: config.engine.caching.archive.archiveFolder,
+      retentionDuration: 8760,
```
burgerni10, line 493: That means that someone who update OIBus with an activated archive mode will lose its data older than one year. It is a breaking change compared to the old archive management (where everything is kept forever). 
We may add another behaviour : after the migration, the new value will be 0 (instead of one year). The input in the front should accept a minimum value of 0 and tell in the help that 0 stands for 'infinite retention period'.
In the backend, you must check if the rettentionDuration is greater than 0 before activating a deletion cycle.

tests/testConfig.js
```diff
@@ -24,7 +24,14 @@ const testConfig = {
         host: 'localhost:3100',
       },
     },
-    caching: { cacheFolder: './cache', archiveMode: 'delete', archiveFolder: './cache/archived/' },
+    caching: { 
+      cacheFolder: './cache', 
+      archive: {
+        enabled: true,
+        archiveFolder: "./cache/archived/",
```
burgerni10, line 31: "archived' -> archive

burgerni10 (changes requested):
The file Cache.class.js has no test. Can you add tests for this class in this PR ?

src/engine/Cache.class.js
```diff
@@ -6,6 +6,8 @@ const Queue = require('../services/queue.class')
 const Logger = require('./Logger.class')
 const ApiHandler = require('../north/ApiHandler.class')
 
+const ARCHIVE_TIME_OUT = 3600000
```
burgerni10, line 9: ARCHIVE_TIME_OUT -> ARCHIVE_TIMEOUT

src/client/Engine/Caching.jsx
```diff
@@ -66,9 +66,9 @@ const Caching = ({ onChange, caching }) => (
           name="engine.caching.archive.retentionDuration"
           label="Retention Duration"
           value={caching.archive.retentionDuration}
-          defaultValue={8760}
+          defaultValue={720}
           valid={validation.engine.caching.archive.retentionDuration}
-          help={<div>How long to wait for the files to get delete in the archive folder (in hours)</div>}
+          help={<div>Retention period of archived files (in hours)</div>}
```
burgerni10, line 71: Can you update the main 'help' section inside the OIbTitle ?
Please check that it's coherent with the new fields names and add the retentionDuration documentation (especially about the 0 value that means the files will be kept forever)

burgerni10 (changes requested):

PR #1335, opened 2021-07-29
fix(test): unit test for FileWriter 

franckfotso1:
Capture d’écran 2021-07-29 à 14 38 53


burgerni10 (changes requested):
You can check the behaviour of the connector when errors are thrown (l43-44 and l64-65).

Let me know if you need help to simulate an error. 

Here is a hint: you can do that by mocking a package method that throws an error instead of doing nothing

franckfotso1:
Capture d’écran 2021-08-03 à 10 13 01

PR #1347, opened 2021-08-11
Bulk history query

kukukk:


src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -63,14 +66,21 @@ class SQLDbToFile extends ProtocolHandler {
...
     this.dateFormat = dateFormat
     this.compression = compression
+    this.maxReadInterval = maxReadInterval
+    this.readIntervalDelay = readIntervalDelay
+    this.maxReturnValues = maxReturnValues
+
+    const hasStartTime = this.query.indexOf('@StartTime') === -1 ? 0 : 1
+    const hasEndTime = this.query.indexOf('@EndTime') === -1 ? 0 : 1
+    const hasMaxReturnValues = this.query.indexOf('@MaxReturnValues') === -1 ? 0 : 1
+    this.nrOfReplacements = hasStartTime + hasEndTime + hasMaxReturnValues
```
burgerni10, line 83: nrOfReplacements -> numberOfReplacements

src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -164,8 +192,9 @@ class SQLDbToFile extends ProtocolHandler {
     this.logger.debug(`Found ${result.length} results`)
 
     if (result.length > 0) {
-      this.lastCompletedAt[scanMode] = this.setLastCompletedAt(result, this.lastCompletedAt[scanMode])
+      this.lastCompletedAt[scanMode] = this.setLastCompletedAt(result, startTime)
       await this.setConfig(`astCompletedAt-${scanMode}`, this.lastCompletedAt[scanMode].toISOString())
```
burgerni10, line 196: Missing the L from lastCompletedAt

src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -204,12 +233,13 @@ class SQLDbToFile extends ProtocolHandler {
...
-   * @returns {void}
+   * @param {Date} startTime - The start time
+   * @param {Date} endTime - The end time
+   * @returns {Promise<array>} - The new entries
    */
-  async getDataFromMSSQL(lastCompletedAt) {
+  async getDataFromMSSQL(startTime, endTime) {
     const adaptedQuery = this.query
-    this.logger.debug(`Executing "${adaptedQuery}" ${this.containsLastCompletedDate ? 'with' : 'without'} LastCompletedDate`)
+    this.logger.debug(`Executing "${adaptedQuery}" ${this.nrOfReplacements ? 'with' : 'without'} StartTime/EndTime/MaxReturnValues`)
```
burgerni10, line 242: Can you replace this log by a more elaborate one ? Something like :
Executing "${adaptedQuery}" with StartTime = ${StartTime.toISOString()} (string formatted),  EndTime = ${EndTime.toISOString()} (string formatted), MaxReturnValues = ${MaxReturnValues}

Displaying only the variable used in the query.

src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -228,9 +258,11 @@ class SQLDbToFile extends ProtocolHandler {
     try {
       const pool = await new mssql.ConnectionPool(config).connect()
       let result
-      if (this.containsLastCompletedDate) {
+      if (this.nrOfReplacements) {
         result = await pool.request()
-          .input('LastCompletedDate', mssql.DateTimeOffset, lastCompletedAt)
+          .input('StartTime', mssql.DateTimeOffset, startTime)
+          .input('EndTime', mssql.DateTimeOffset, endTime)
+          .input('MaxReturnValues', mssql.Numeric, this.maxReturnValues)
```
burgerni10, line 265: We noticed that DateTimeOffset does not support date operation in MSSQL queries. Is there any reason why to use that instead of DateTime ?

src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -270,7 +303,7 @@ class SQLDbToFile extends ProtocolHandler {
     let data = []
     try {
       connection = await mysql.createConnection(config)
-      const params = this.containsLastCompletedDate ? [lastCompletedAt] : []
+      const params = this.nrOfReplacements ? [startTime, endTime, this.maxReturnValues] : []
```
burgerni10, line 306: Need to check how it behaves if endTime is not used but maxReturnValues is used

burgerni10: Same remarks for other divers

src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -101,35 +121,44 @@ class SQLDbToFile extends ProtocolHandler {
...
       this.logger.error('Invalid timezone')
       return
     }
 
-    if (this.ongoingReads[scanMode]) {
-      this.logger.silly(`onScan ignored: ongoingReads[${scanMode}]: ${this.ongoingReads[scanMode]}`)
+    if (!((this.nrOfReplacements === 0) || (this.nrOfReplacements === 3))) {
+      this.logger.error('Invalid query format. Please use all or nothing from @StartTime, @EndTime, @MaxReturnValues')
       return
     }
```
burgerni10, line 163: After we discussed about the migration steps, I think we should allow to use only some of these variables. Particularly StartTime only so it will be easier to automatically migrate LastCompletedDate to StartTime

burgerni10 (changes requested):
I've noticed some conflicts with past rebase of the bulk/develop branch. I'll fix it when merging and I will update the bulk/develop branch after this PR will be merged

src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -429,6 +476,58 @@ class SQLDbToFile extends ProtocolHandler {
...
+  }
+
+  /**
+   * Get all occurrences of a substring with a value
+   * @param {string} str - The string to look for occurrences in
+   * @param {string} keyword - The keyword
+   * @param {any} value - The value to assign to the occurrences index
+   * @return {object[]} - The result as { index, value}
+   */
+  static getOccurrences(str, keyword, value) {
```
burgerni10, line 519: Can you add specific tests and examples for this method like you did for formatDateWithTimezone method ?

src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -429,6 +476,58 @@ class SQLDbToFile extends ProtocolHandler {
...
+
+  /**
+   * Generate replacements parameters
+   * @param {string} query - The query
+   * @param {Date} startTime - The StartTime
+   * @param {Date} endTime - The EndTime
+   * @param {number} maxReturnValues - The MaxReturnValues
+   * @return {any[]} - The replacement parameters
+   */
+  static generateReplacementParameters(query, startTime, endTime, maxReturnValues) {
```
burgerni10, line 503: Can you add specific tests and examples for this method like you did for formatDateWithTimezone method ?

src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -91,81 +96,100 @@ class SQLDbToFile extends ProtocolHandler {
...
+        result = result.concat(intervalResult)
+        // eslint-disable-next-line no-param-reassign
+        startTime = this.getLatestDate(result, startTime)
+      } catch (error) {
+        this.logger.error(error)
       }
-    } catch (error) {
-      this.logger.error(error)
-    }
+    } while (intervalResult.length)
```
burgerni10, line 188: Instead of this check, wouldn't it be better to check intervalResult.length AND the last value/timestamp returned from the previous intervalResult ? If both are the same, it will leave the while loop and avoid the problem you saw in PostgreSQL

burgerni10 (changes requested):

burgerni10 (approved):

jfhenon (changes requested):

src/south/SQLDbToFile/SQLDbToFile.schema.jsx
```diff
@@ -16,7 +16,9 @@ schema.form = {
         </p>
         <ul>
           <li>
-            The query may have a specific format and contain a WHERE clause with the date constraint of the time column using @LastCompletedDate.
+            The query may have a specific format and contain a WHERE clause with the date constraint of the time column using @StartTime and @EndTime.
+            It can also contain a LIMIT clause with @MaxReturnValues to limit the number of results.
+            Note: The query should contain either all parameters all none of them.
```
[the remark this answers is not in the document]
kukukk: fixed

src/south/ProtocolHandler.class.js
```diff
@@ -132,9 +134,52 @@ class ProtocolHandler {
     this.logger.info(`Data source ${dataSourceId} started with protocol ${protocol}`)
   }
 
-  onScanImplementation(scanMode) {
-    const { dataSourceId } = this.dataSource
-    this.logger.error(`Data source ${dataSourceId} should surcharge onScanImplementation(${scanMode})`)
+  async onScanImplementation(scanMode) {
```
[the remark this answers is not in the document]
kukukk: Indeed, it is not the best naming. `historyQueryImplementation` would be ok?

src/south/OPCHDA/OPCHDA.class.js
```diff
@@ -50,10 +54,17 @@ class OPCHDA extends ProtocolHandler {
...
    * @return {Promise<void>} - The on scan promise
    */
-  onScanImplementation(scanMode) {
-    this.sendReadMessage(scanMode)
+  async historyQuery(scanMode, startTime, endTime) {
+    this.sendReadMessage(scanMode, startTime, endTime)
+    // Wait until we receive the response
+    while (this.ongoingReads[scanMode]) {
+      // eslint-disable-next-line no-await-in-loop
+      await this.delay(100)
```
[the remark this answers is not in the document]
kukukk: For `OPCUA-HA` and `SQLDbToFile` the `historyQuery` implementation returns a promise, so in `ProtolHandler` the `await historyQuery()` call will wait until the response is received. In case of `OPCHDA` the `sendReadMessage` is not returning a promise, so it will immediately return before we receive the response from the agent. The flag for `ongoingReads[scanMode]` is reset after we receive the response from the agent. The `while` loop waits until the response is received.

src/south/ProtocolHandler.class.js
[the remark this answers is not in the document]
jfhenon: Implementation tends to indicate that it is in the child class where you do specific implementation (as it was before). 
Now, it's a part of parentClass so maybe historyQueryHandler or historyQueryHandler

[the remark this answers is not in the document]
kukukk: historyQueryHandler sounds good.

src/south/OPCHDA/OPCHDA.class.js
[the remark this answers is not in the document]
jfhenon: I understand how this works but I would like to rethink this. I think we should await a promise and the response should resolve it. 
However, it's not really related to this PR so we could open a new issue for OPCHDA

src/south/ProtocolHandler.class.js
```diff
@@ -134,7 +134,7 @@ class ProtocolHandler {
     this.logger.info(`Data source ${dataSourceId} started with protocol ${protocol}`)
   }
 
-  async onScanImplementation(scanMode) {
+  async git ad(scanMode) {
```
jfhenon, line 137: ???

kukukk: Sorry, I didn't realized that IntelliJ was the active window instead of Terminal when I started to type `git add`. I fixed it.

src/south/OPCHDA/OPCHDA.class.js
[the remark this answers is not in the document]
kukukk: Something like this? https://stackoverflow.com/a/36072263/436312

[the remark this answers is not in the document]
jfhenon: yes, seems like a good example. (but let's treat this as another issue)

burgerni10 (approved):

PR #1355, opened 2021-08-20
fix(north-south): Use permanent id field for database file name

dezse97:
Solves #1191 

- generate `uuid` for new applications and datasources (frontend side)
- save it to config file
- migration for existing cache file to avoid data loss
- updated the tests

### DIFF
--- a/src/client/Engine/ExternalSources.jsx
+++ b/src/client/Engine/ExternalSources.jsx
@@ -23,7 +23,7 @@ const ExternalSources = ({ externalSources }) => {
         <OIbTitle label="ExternalSources">
           <p>
             This is a list of dataSources from another OIBus. The name should respect the following convention:
-            OIbusId:name
+            OIBusId:DataSourceName
           </p>
         </OIbTitle>
         <Row>
--- a/src/client/Engine/ExternalSources.jsx
+++ b/src/client/Engine/ExternalSources.jsx
@@ -23,7 +23,7 @@ const ExternalSources = ({ externalSources }) => {
         <OIbTitle label="ExternalSources">
           <p>
             This is a list of dataSources from another OIBus. The name should respect the following convention:
-            OIbusId:datasourceId
+            OIbusId:name
           </p>
         </OIbTitle>
         <Row>
--- a/src/client/North/Form/NorthForm.jsx
+++ b/src/client/North/Form/NorthForm.jsx
@@ -10,7 +10,7 @@ import OIbForm from '../../components/OIbForm/OIbForm.jsx'
 import ApiSchemas from '../Apis.jsx'
 
 const NorthForm = ({ application, applicationIndex, onChange }) => {
-  const { api, applicationId } = application
+  const { id, api, applicationId } = application
   // Create the sections for the api (for example application.Link) for application not yet initialized
   if (!application[api]) application[api] = {}
   if (!application.caching) application.caching = {}
@@ -31,6 +31,7 @@ const NorthForm = ({ application, applicationIndex, onChange }) => {
           </BreadcrumbItem>
           <BreadcrumbItem active tag="span">
             {applicationId}
+            <small>{` (${id || ''})`}</small>
           </BreadcrumbItem>
         </Breadcrumb>
       </Row>
--- a/src/client/North/Form/NorthForm.jsx
+++ b/src/client/North/Form/NorthForm.jsx
@@ -10,7 +10,7 @@ import OIbForm from '../../components/OIbForm/OIbForm.jsx'
 import ApiSchemas from '../Apis.jsx'
 
 const NorthForm = ({ application, applicationIndex, onChange }) => {
-  const { id, api, applicationId } = application
+  const { api, name } = application
   // Create the sections for the api (for example application.Link) for application not yet initialized
   if (!application[api]) application[api] = {}
   if (!application.caching) application.caching = {}
@@ -30,8 +30,7 @@ const NorthForm = ({ application, applicationIndex, onChange }) => {
             North
           </BreadcrumbItem>
           <BreadcrumbItem active tag="span">
-            {applicationId}
-            <small>{` (${id || ''})`}</small>
+            {name}
           </BreadcrumbItem>
         </Breadcrumb>
       </Row>
--- a/src/client/North/NewApplicationRow.jsx
+++ b/src/client/North/NewApplicationRow.jsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { Button, Form, Col, Row } from 'reactstrap'
+import { v4 as uuidv4 } from 'uuid'
 import { OIbText, OIbSelect } from '../components/OIbForm/index'
 
 const NewApplicationRow = ({ apiList, addApplication }) => {
@@ -14,7 +15,7 @@ const NewApplicationRow = ({ apiList, addApplication }) => {
   const handleAddApplication = () => {
     //  update the new application's state
     if (applicationId === '') return
-    addApplication({ applicationId, api })
+    addApplication({ id: uuidv4(), applicationId, api })
   }
 
   const handleChange = (name, value) => {
--- a/src/client/North/NewApplicationRow.jsx
+++ b/src/client/North/NewApplicationRow.jsx
@@ -1,11 +1,11 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { Button, Form, Col, Row } from 'reactstrap'
-import { v4 as uuidv4 } from 'uuid'
+import { nanoid } from 'nanoid'
 import { OIbText, OIbSelect } from '../components/OIbForm/index'
 
 const NewApplicationRow = ({ apiList, addApplication }) => {
-  const [applicationId, setApplicationId] = React.useState('')
+  const [name, setName] = React.useState('')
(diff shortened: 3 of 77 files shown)


src/client/North/Form/NorthForm.jsx
```diff
@@ -10,7 +10,7 @@ import OIbForm from '../../components/OIbForm/OIbForm.jsx'
 import ApiSchemas from '../Apis.jsx'
 
 const NorthForm = ({ application, applicationIndex, onChange }) => {
-  const { api, applicationId } = application
+  const { id, api, applicationId } = application
   // Create the sections for the api (for example application.Link) for application not yet initialized
```
jfhenon, line 14: if the id is given, do we need to pass also the applicationId? 
In term of naming, the applicationId should likely be applicationName and the id should be applicationId

src/client/North/Form/NorthForm.jsx
```diff
@@ -31,6 +31,7 @@ const NorthForm = ({ application, applicationIndex, onChange }) => {
           </BreadcrumbItem>
           <BreadcrumbItem active tag="span">
             {applicationId}
+            <small>{` (${id || ''})`}</small>
           </BreadcrumbItem>
```
jfhenon, line 35: should the 'id' visible to end users?

src/client/North/NewApplicationRow.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { Button, Form, Col, Row } from 'reactstrap'
+import { v4 as uuidv4 } from 'uuid'
```
jfhenon, line 4: just for info.  I have seen a new package called nanoid for generatiing unique ids.

src/client/North/North.jsx
```diff
@@ -77,11 +77,11 @@ const North = () => {
    * applicationId, enabled and api fields
    * @returns {void}
    */
-  const addApplication = ({ applicationId, api }) => {
+  const addApplication = ({ id, applicationId, api }) => {
     const applicationIndex = getApplicationIndex(applicationId)
```
jfhenon, line 81: same comment about the id/applicationId

src/client/South/Form/SouthForm.jsx
```diff
@@ -9,7 +9,7 @@ import PointsButton from '../PointsButton.jsx'
 import StatusButton from '../StatusButton.jsx'
 
 const SouthForm = ({ dataSource, dataSourceIndex, onChange }) => {
-  const { protocol, dataSourceId } = dataSource
+  const { id, protocol, dataSourceId } = dataSource
   const history = useHistory()
```
jfhenon, line 13: same comment about id and dataSourceId versus dataSourceId, dataSourceName

src/client/North/Form/NorthForm.jsx
dezse97: In this case the `applicationId` is used for the breadcrumb, to see which north application are we editing. Regarding the naming, yes, I think so. But it would be a huge changeset to rename all the occurrences in the project. Should I apply those changes?

src/client/North/NewApplicationRow.jsx
dezse97: Ok, thanks. As i read, it's faster than uuid and has a smaller size with no additional dependencies. I am going to switch the library

src/client/North/Form/NorthForm.jsx
dezse97: I thought it's not a bad idea, but we can show only the name. Anyway, in the config file the id will be visible, if we need it for some reason

src/client/North/Form/NorthForm.jsx
jfhenon: I leave it @burgerni10 to decide.

src/client/North/Form/NorthForm.jsx
jfhenon: maybe it should be only visible when "hovering" on the name.

kukukk: If I remember correctly, we talked about showing it on the UI to make easier identifying the database file, in case we need the client to send it to us for debug/support.

dezse97: Yes, I implemented based on the previous comments on the related issue. Maybe we can remove the id field from the table, but leave it in the breadcrumb, so the user will be able to copy from there if needed

jfhenon: In fact, we have a "status" screen for each application/protocol. I think the uuid would be useful there.

src/client/North/Form/NorthForm.jsx
burgerni10: I think we can keep "id" like that and rename  applicationId to "name" to have consistency across south and north connectors.
You can apply those changes in the whole code.

src/client/North/Form/NorthForm.jsx
burgerni10: Indeed, the applicationId/name should be visible in case we need to access the .db file which will have the id in its filename. I agree that this id is better inside the status page

src/client/North/Form/NorthForm.jsx
dezse97: Ok. So both `applicationId` and `dataSourceId` would be `name`? Wouldn't be cleaner if we leave the application and dataSource keyword there (like `applicationName` and `dataSourceName` suggested by @jfhenon )?. The data structure is different anyway and would be easier to make difference between north and south names in the code.

burgerni10: I think id and name are better since they have the same purposes for both South and North connectors. We usually name the north connector "application" and the south connector "dataSource" anyway. So it would still be clear in the code since we will have application.name or dataSource.id

dezse97: I see, anyway it is a clean coding pattern to not have redundant attribute names, that can be found out from the main object name. Alright, I'll proceed

burgerni10 (changes requested):

src/engine/Engine.class.spec.js
```diff
@@ -109,7 +109,7 @@ describe('Engine', () => {
     }]
 
     engine.cache.cacheValues = jest.fn()
-    await engine.addValues('sourceId', sampleValues)
+    await engine.addValues('sourceId', 'sourceName', sampleValues)
```
jfhenon, line 112: It should be possible to deduct the name from the id? why do we give the 2 parameters?

src/client/Engine/ExternalSources.jsx
```diff
@@ -23,7 +23,7 @@ const ExternalSources = ({ externalSources }) => {
         <OIbTitle label="ExternalSources">
           <p>
             This is a list of dataSources from another OIBus. The name should respect the following convention:
-            OIbusId:datasourceId
+            OIbusId:name
```
jfhenon, line 26: We are explaining to users so maybe here OIbusId:SourceName maybe a better explanation. of <<OIBUS id>>:<<Source Name>> for example 'OIBus1:HDA'

src/client/South/SouthStatus.jsx
```diff
@@ -9,15 +9,15 @@ import { AlertContext } from '../context/AlertContext.jsx'
...
+  const { id } = useParams() // the dataSource id passed in the url
 
   /**
    * Acquire the status
    * @returns {void}
    */
   const fetchStatus = () => {
     apis
-      .getSouthStatus(dataSourceId)
+      .getSouthStatus(id)
```
jfhenon, line 20: Even if the name is retrieved in the status object, should the getSouthStatus ask for the name or the internal id ?

src/client/index.jsx
```diff
@@ -34,11 +34,11 @@ const Main = () => (
               <Route exact path="/" component={Health} />
               <Route exact path="/engine" component={Engine} />
               <Route exact path="/south" component={South} />
-              <Route exact path="/south/:dataSourceId" component={ConfigureProtocol} />
-              <Route exact path="/south/:dataSourceId/live" component={SouthStatus} />
-              <Route exact path="/south/:dataSourceId/points" component={ConfigurePoints} />
+              <Route exact path="/south/:id" component={ConfigureProtocol} />
```
jfhenon, line 37: same question. should the id or the name be visible in the path

jfhenon (commented):
Some comments. 
The main point is to always think twice if we work with the id or the name.

src/client/South/SouthStatus.jsx
burgerni10: Everything used to identify the connector (south or north) should be the id, and name should be used for display only.

src/client/index.jsx
burgerni10: The url is used to identify the connector, hence the use of the id instead of the name. Moreover, with an Id in the url, we have the same pattern than OIAnalytics to identify entities (dashboards, users,...)

src/engine/Engine.class.spec.js
burgerni10: Indeed, we could retrieve it from the "activeApis" or "activeProtocols" lists from the engine

jfhenon (approved):
verifier/corriger status.Name

jfhenon (approved):

PR #1359, opened 2021-08-21
Cache and debounce

jfhenon:
Debounce: when a protocol receives many calls within a second, the engine is called multiple times (i.e. causing sqlite to be called inefficiently). This was causing an issue for a customer having several hundreds calls per second with a MQTT source. 
This PR proposes to group all calls within 300ms or below 250 values to be buffered and grouped into a single call to the engine. The 300ms/250 are now hard coded values but we may consider to expose them to the UI. 
Note that a single value will be delayed by 300ms before being sent to the engine.

### DIFF
--- a/src/south/ProtocolHandler.class.js
+++ b/src/south/ProtocolHandler.class.js
@@ -105,18 +105,26 @@ class ProtocolHandler {
   }
 
   /**
+   * Method used to flush the buffer from a time trigger or a max trigger
+   * @param {string} flag - The trigger
+   * @returns {void}
+   */
+  async flush(flag = 'time') {
+    this.logger.silly(`${flag}: ${this.buffer.length}, ${this.dataSource.dataSourceId}`)
+    await this.engine.addValues(this.dataSource.dataSourceId, this.buffer)
+    this.buffer = []
+    if (this.bufferTimeout) {
+      clearTimeout(this.bufferTimeout)
+      this.bufferTimeout = null
+    }
+  }
+
+  /**
    * Add a new Value to the Engine.
    * @param {array} values - The new value
    * @return {void}
    */
-  addValues(values) {
-    const flush = async (flag = 'time') => {
-      this.logger.silly(`${flag}: ${this.buffer.length}, ${this.dataSource.dataSourceId}`)
-      await this.engine.addValues(this.dataSource.dataSourceId, this.buffer)
-      this.buffer = []
-      clearTimeout(this.bufferTimeout)
-      this.bufferTimeout = null
-    }
+  async addValues(values) {
     // used for status
     this.lastAddPointsAt = new Date().getTime()
     this.addPointsCount += values.length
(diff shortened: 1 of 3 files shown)


jfhenon:
note that ADS tests have been impacted (calls were grouped) and has been adjusted.

jfhenon:
In the same PR, I fixed an issue with the cache that was checking the archive folder even if archive was not enabled.

kukukk:
I think it can lead to data loss. If OIBus is stoppped after the South calls `addValues` but before the buffer is flushed and `this.engine.addValues` is called. To avoid this we should save lastCompletedAt in the database only when we flush the buffer.

src/south/ProtocolHandler.class.js
```diff
@@ -101,9 +110,26 @@ class ProtocolHandler {
    * @return {void}
    */
   addValues(values) {
+    const flush = async (flag = 'time') => {
+      this.logger.silly(`${flag}: ${this.buffer.length}, ${this.dataSource.dataSourceId}`)
+      await this.engine.addValues(this.dataSource.dataSourceId, this.buffer)
+      this.buffer = []
+      clearTimeout(this.bufferTimeout)
```
burgerni10, line 117: Need to check if bufferTimeout is null. Can happen if more than BUFFER_MAX values are retrieved at first call of this function

src/south/ProtocolHandler.class.js
```diff
@@ -101,9 +110,26 @@ class ProtocolHandler {
...
+      this.bufferTimeout = null
+    }
```
burgerni10, line 119: This method can be declared as private ProtocolHandler method so it won't create it at every addValues calls

src/south/ProtocolHandler.class.js
```diff
@@ -101,9 +110,26 @@ class ProtocolHandler {
    * @return {void}
    */
   addValues(values) {
```
burgerni10, line 112: As Szabi mentionned, we can have data loss. Maybe we can pass the lastCompletedAt and the dataSourceId as parameters, store them in an array and update the database at the end of the flush to fix that

src/south/ProtocolHandler.class.js
```diff
@@ -101,9 +110,26 @@ class ProtocolHandler {
...
     this.lastAddPointsAt = new Date().getTime()
     this.addPointsCount += values.length
-    this.engine.addValues(this.dataSource.dataSourceId, values)
+    // add new values to the protocol buffer
+    this.buffer.push(...values)
+    // if the protocol buffer is large enough, send it
+    // else start a timer before sending it
+    this.logger.silly(`${this.buffer.length}, ${!!this.bufferTimeout}, ${this.dataSource.dataSourceId}`)
+    if (this.buffer.length > BUFFER_MAX) {
+      flush('max')
```
burgerni10, line 129: Shouldn't we await this method to avoir concurrency access to the buffer ?

burgerni10 (changes requested):

PR #1376, opened 2021-09-05
fix(engine): add bufferMax and bufferTimeoutInterval used for points protocols

burgerni10:


burgerni10:
I also correct MQTT protocol settings typo for consistency.
There was also an error in the Cache when archive mode was disabled

PR #1383, opened 2021-09-10
feat(north): Allow value as json in data entry for some North connectors

Mollier-Pierret:
new branch from develop including proposal code to allow value as json in data entry for InfluxDB, MongoDB, TimeScaleDB, MQTTNorth connectors


burgerni10:
I've added som test cases and adjust the code from what I understood. @Mollier-Pierret can you test this code in your use cases ? Please take a look at my unit tests to see if it matches yours. I've also updated this branch with the many changes of the develop branch, you may need to reinstall packages and start with a new oibus.json.

On the code itself, I think we can factorise some of it like the getJsonValueByStringPath and escapeSpace utility functions. 
@jfhenon Can you take a look at the PR and give feedbacks ? Issues #1412, #1413 and #1384 should be fixed with this PR

src/north/influxdb/InfluxDB.class.js
```diff
@@ -2,6 +2,25 @@ const { vsprintf } = require('sprintf-js')
...
 const ApiHandler = require('../ApiHandler.class')
 
+/**
+ * function return the content of value, that could be a Json object with path keys given by string value
+ * without using eval function
+ * @param {*} value - simple value (integer or float or string, ...) or Json object
+ * @param {*} pathValue - The string path of value we want to retrieve in the Json Object
+ * @return {*} The content of value depending on value type (object or simple value)
+ */
+const getJsonValueByStringPath = (value, pathValue) => {
```
jfhenon, line 12: I would give an example of the output of this function.
Also, we are using a library called ""object-path" already and I feel it can provide the same result.

PR #1385, opened 2021-09-10
feat(mqtt): allow mqtts authenticate with certificates

Mollier-Pierret:
new branch from develop including proposal code to allow mqtts authenticate with certificates for South connector MQTT and North connector MQTTNorth


jfhenon:
I think this is a very good improvement. 

We use a code almost identical in the north and the south extension for MQTT connect, could we find a way to factor this code?

jfhenon (approved):

Issue #1380, opened 2021-09-06
[UI - Point list] - Remove the wrong point when the list is filtered


burgerni10:
**Describe the bug**
On every point protocols, on the point list, when the list is filtered, the deletion remove the wrong point

**To Reproduce**
1. Create a point connector (mqtt, ads...)
2. Go the the point list and add some points
3. Filter your list by entering something in the filter and make it so it remove some points of the list
4. Remove a point (the removed point index of the list is used, hence the error)

dezse97:
I would extend this issue with the "edit" functionality, as it has the same behavior too.

PR #1398, opened 2021-09-20
fix(points-filter): Fix "remove" and "edit" on filtered point list

dezse97:
The `reverseIndex` is not a good approach, as the ids in the filtered list are not the same (there are fewer items) so the deletion and update will operate with a wrong id.
Solved it by using a virtual index on the frontend side which is constant even in the filtered list ensuring that the chosen point will be edited or removed.

Solves #1380

### DIFF
--- a/src/client/South/ConfigurePoints.jsx
+++ b/src/client/South/ConfigurePoints.jsx
@@ -41,15 +41,22 @@ const ConfigurePoints = () => {
   const { points: pointsOrdered = [], protocol } = dataSource
   const points = pointsOrdered.slice().reverse()
 
+  // filter
+  const filteredPoints = filterText
+    ? points.filter(
+      (point) => Object.values(point).findIndex((element) => element
+        .toString()
+        .toLowerCase()
+        .includes(filterText.toLowerCase())) >= 0,
+    ) : points
+
   /**
-   * Get the index of point for reversed poitns list
-   * @param {number} index of point
-   * @returns {number} return reversed index
+   * @param {number} index the index of a point in the table
+   * @returns {number} the index in the config file of the chosen point
    */
-  const reversedIndex = (index) => {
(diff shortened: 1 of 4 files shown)


src/client/South/ConfigurePoints.jsx
```diff
@@ -39,19 +39,44 @@ const ConfigurePoints = () => {
...
   /**
-   * Get the index of point for reversed poitns list
-   * @param {number} index of point
-   * @returns {number} return reversed index
-   */
+  * Get the index of point for reversed poitns list
+  * @param {number} index of point
+  * @returns {number} return reversed index
+  */
   const reversedIndex = (index) => {
```
burgerni10, line 64: This method is not used elsewhere. I think it can be merged with findIndexBasedOnVirtualIndex ?

src/client/South/ConfigurePoints.jsx
```diff
@@ -133,7 +158,7 @@ const ConfigurePoints = () => {
   const onChange = (name, value, validity) => {
     // add pageOffet before dispatch the update to update the correct point (pagination)
```
burgerni10, line 159: pageOffet -> pageOffset

burgerni10 (changes requested):
Minor comments, otherwise the logic seems ok. Can you improve the test coverage to get 100% coverage on this file ?
Current coverage is : 
File                              | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
  ConfigurePoints.jsx             |   90.72 |    80.95 |    87.5 |   92.05 | 105,193-198

jfhenon (approved):

src/client/South/ConfigurePoints.jsx
dezse97: It can be merged, but i left a comment to be clear what's happening in those lines

Issue #1053, opened 2021-01-21
[South - OPCUA] - Manage certificates


burgerni10:
Allow authentication with cert files : the user must specify the path to the cert files to be used.

PR #1406, opened 2021-09-29
Opcua certificates

burgerni10:
ON TOP OF #1385 

Work in Progress branch. Node-opcua documentation is laking on certificate usage. I need an opcua server which manage security policies and client authentication with certificate to test and improve this work.


burgerni10:
Issue #1053

src/south/OPCUA_DA/OPCUA_DA.class.js
```diff
@@ -162,22 +179,59 @@ class OPCUA_DA extends ProtocolHandler {
     this.reconnectTimeout = null
 
     try {
+      let keyFileContent = ''
```
jfhenon, line 182: It seems that the the same very similar code is used to manage certificates in different South. 
Should we move this to you he Protocol class or a library?

burgerni10: Those files are the same for OPCUA_DA / HA, but they differ on MQTT (where it needs three files for example). Some additional checks for the format of the files may differ in the future depending of the protocol settings.

That's why I put the logic inside the data source anyway.

jfhenon (approved):

Issue #1428, opened 2021-10-13
Hide console window when starting OIBus manually under Windows OS


Mollier-Pierret:
When we start manually (i.e. not using services) OIBus under Windows OS, in Node.js server application, using spawn with parameter {windowsHide:true} the console associated to Master process doesn't appear, but this associated to Worker process appear.

It's possible to hide the console associated to Worker process by adding a line "cluster.setupMaster({ windowsHide: true })" before executing "cluster.fork()"

Could it be possible to integrate this functionnality in future OIBus release ?

PR #1472, opened 2021-11-10
Windows installer fix

S-Canler:


deps/win/oibus_setup.iss
```diff
@@ -271,9 +272,9 @@ begin
     if FileExists(JsonFile) then
     begin
       ConfExists := True;
-      if MsgBox('An oibus.json file was found at ' + MyDataDir + '. Do you want to overwrite it?', mbInformation, MB_YESNO) = IDYES then
+      if MsgBox('An oibus.json file was found at ' + MyDataDir + '. Do you want to use it for your OIBus session ?', mbInformation, MB_YESNO) = IDNO then
       begin
-        if MsgBox('WARNING : Overwriting the current setup will delete all logins, passwords and data you saved so far.' + #13#10 + 'Are you sure you want to proceed?', mbInformation, MB_YESNO) = IDNO then
+        if MsgBox('WARNING : You are about to overwrite the current setup and delete all logins, passwords and data you saved so far.' + #13#10 + 'Are you sure you want to proceed ?', mbInformation, MB_YESNO) = IDNO then
```
burgerni10, line 277: The aim is to have the default behaviour not to overwrite. In this case, keeping the origine question is ok if IDNO means that the default selected answer is NO.
Here you double negate the question which ultimately give the same behaviour than before.

deps/win/oibus_setup.iss
```diff
@@ -237,10 +238,10 @@ begin;
             Result := False
           else
           begin
-            FileContent := 'echo Stopping OIBus service...' + #13#10
-              + 'nssm.exe stop OIBus >nul 2>&1' + #13#10
+            FileContent := 'echo Stopping OIBus service... ' + #13#10
+              + 'nssm.exe stop OIBus >nul 2>&1 &> /dev/null' + #13#10
               + '@echo Starting OIBus in the console...' + #13#10
-              + '"' + ExpandConstant('{app}') + '\oibus.exe" --config "' + ConfigFilePath + '"'
+              + '"' + ExpandConstant('{app}') + '\oibus.exe" --config "' + ConfigFilePath + '" &> /dev/null'
```
burgerni10, line 244: I told you to do that in a separate PR because it focus on something different that I don't want to test this feature in this PR. 
Moreover, from what I see, you do not offer the user the 'hideConsole' as an option, it seems always hidden. Lastly this seems to hide messages but the terminal itself seems to stay open which is not what the issue #1428  is asking.

burgerni10 (changes requested):
I cannot accept this PR because of the comments but also because of the 8 commits I see in the PR : why don't you have only one commit since you only edit this file ? 
Please refresh this PR or redo it from scratch to keep it clean

burgerni10 (approved):

PR #1493, opened 2021-11-22
feat(bulk): Logic to iterate HistoryQuery and export data

kukukk:


src/HistoryQuery/HistoryQuery.class.js
```diff
@@ -9,51 +9,172 @@ class HistoryQuery {
...
+    // }
+
+    // const { api, enabled, name } = this.application
+    // this.north = enabled ? ApiFactory.create(api, this.application, this.engine) : null
+    // if (this.north) {
+    //   await this.north.connect()
+    // } else {
+    //   this.logger.error(`Application ${name} is enabled or not found`)
+    //   this.engine.runNextHistoryQuery()
+    // }
```
burgerni10, line 107: Is there a reason it is commented ?

src/HistoryQuery/HistoryQueryEngine.class.js
```diff
@@ -55,15 +75,103 @@ class Engine extends BaseEngine {
...
+
+  /**
+   * Get the next history query config
+   * @returns {object|null} - The next history query config
+   */
+  getNextHistoryQueryConfig() {
+    const historyQueryConfigs = this.configService.getActiveHistoryQueryConfiguration()
+    const activeHistoryQueryConfigs = historyQueryConfigs.filter((historyQueryConfig) => historyQueryConfig.enabled && !historyQueryConfig.paused)
+
+    const exportingHistoryQueries = activeHistoryQueryConfigs.filter(
```
burgerni10, line 104: This variable should be renamed since it includes the "exporting" and "importing" status
I suggest ongoingHistoryQueries for example

burgerni10 (changes requested):
Small changes, otherwise the logic looks good to me !

PR #1528, opened 2022-01-11
feat(dependencies): Migrate from moment to luxon

kukukk:


src/south/SQLDbToFile/SQLDbToFile.class.js
```diff
@@ -121,10 +121,30 @@ class SQLDbToFile extends ProtocolHandler {
   setLastCompletedAt(entryList) {
     let newLastCompletedAt = this.lastCompletedAt
     entryList.forEach((entry) => {
-      if (entry[this.timeColumn] instanceof Date && entry[this.timeColumn] > new Date(newLastCompletedAt)) {
-        newLastCompletedAt = entry[this.timeColumn].toISOString()
-      } else if (entry[this.timeColumn] && new Date(entry[this.timeColumn]).toString() !== 'Invalid Date') {
-        const entryDate = new Date(moment.tz(entry[this.timeColumn], this.timezone).tz('UTC').toISOString())
+      if (entry[this.timeColumn]) {
```
jfhenon, line 124: Switch (true) is not really readable. I would prefer if then else statement.

src/south/SQLDbToFile/SQLDbToFile.class.spec.js
```diff
@@ -533,6 +533,15 @@ describe('SQLDbToFile', () => {
 
   it('should send uncompressed file when the result is not empty and compression is false', async () => {
     const RealDate = Date
+    global.Date = (...args) => {
+      switch (args.length) {
+        case 0:
+          return new RealDate(nowDateString)
+        default:
+          return new RealDate(...args)
```
jfhenon, line 541: Why do we use a switch for one condition?

jfhenon (changes requested):

src/south/SQLDbToFile/SQLDbToFile.class.spec.js
kukukk: Because of copy-paste :} I will rewrite it with if-else.

burgerni10 (approved):

PR #1581, opened 2022-01-25
fix(sql-query): add @QueryParts and ms to @CurrentDate replacement parameters to avoid overwriting generated files of the request

burgerni10:
I've removed the while loop and the MaxReturnedValues in the SQLDbToFile. 
After some tests, I noticed the following problem : introducing MaxReturnedValues yield some bugs that we discussed a little bit in a previous meeting. For example, if we limit the request to 100 values but 1000 values have the same timestamp in the SQL database, we lose 900 values because we incremented the lastCompletedDate to 1 ms. The loop is not useful anymore if I remove the MaxReturnedValues which should be kept only for OPCUA. For OPCUA, MaxReturnedValues is a pernode parameter, and several values for a node and a specific timestamp is not possible with OPC standard

I've override some south parameters in history queries (filename, tmpFolder and a newly created replaceFilenameWithVariable method). 
This way, it fixes errors for SQLDbToFile HistoryQuery where the tmp folder was used instead of the cacheFolder of the History Query. It worked fine for OPCUA because it was managed as points and "intercepted" by the addValues of the HistoryQueryEngine.


src/south/ProtocolHandler.class.js
```diff
@@ -483,6 +491,19 @@ class ProtocolHandler {
     return timestamp
   }
 
+  /**
+   * Replace the variables such as @CurrentDate in the file name with their values
+   * @param {string} filename - The filename to change
+   * @param {number} queryPart - The part of the query being executed
+   * @return {string} newFilename - The renamed file name
+   */
+  replaceFilenameWithVariable(filename, queryPart) {
```
kukukk, line 500: It is very similar to the overwritten version in HistoryQuery. Would not be better to create a `connectorName` South variable initiated with `this.dataSource.name` and overwrite it in HistoryQuery with `${this.dataSource.name}-${this.application.name}`? This way don't have to overwrite this method.

PR #1592, opened 2022-02-02
fix(mqtt): Fix mqtt client ID and improve code quality

burgerni10:


src/engine/Cache.class.js
```diff
@@ -75,7 +75,7 @@ class Cache {
       this.logger.debug(`db count: ${await databaseService.getCount(api.database)}`)
     }
     this.apis[api.id] = api
-    if (api && api.config && api.config.sendInterval) {
+    if (api.config && api.config.sendInterval) {
```
jfhenon, line 78: if (api.config?.sendInterval) ...

jfhenon (approved):

PR #1600, opened 2022-02-07
feat(chore): add CA from Windows user ca folder

burgerni10:


marouanehassanioptimistik (commented):
do we really want to merge into master such confidential package?

PR #1639, opened 2022-03-02
fix(logger): remove logger from services, change silly to trace and add a scope-specific logger for web-server and main-thread

burgerni10:


burgerni10:
@jfhenon I've taken a look at the RequestService (Axios or Fetch) for loki : I don't think it's a good idea to use this one since we use the logger inside the RequestService. 
Moreover, I think we will eventually remove axios and node-fetch dependencies because fetch will be natively included in NodeJS. 

So I suggest we keep the loki logger as it is for now in this PR to avoid big refactoring of the code again.

jfhenon (approved):

jfhenon:
> @jfhenon I've taken a look at the RequestService (Axios or Fetch) for loki : I don't think it's a good idea to use this one since we use the logger inside the RequestService. Moreover, I think we will eventually remove axios and node-fetch dependencies because fetch will be natively included in NodeJS.
> 
> So I suggest we keep the loki logger as it is for now in this PR to avoid big refactoring of the code again.

I agree. We used to propose several libraries when fetch was not supported natively. I think this is not necessary anymore.

PR #1643, opened 2022-03-03
Feat/oibus loki log proxy

### LINKED ISSUE #1325: [Loki logs] - When behind firewalls, make it so that loki logs can be sent to another OIBus
And manage some local forwarding data to display on the oibus

burgerni10:
ON TOP OF #1639 

Resolve #1325


jfhenon (commented):
I will review this one when the silly to trace change will be complete. let me know

src/index.js
```diff
@@ -19,22 +19,23 @@ const MAX_INTERVAL_MILLISECOND = 30 * 1000
...
+
 logger.changeParameters({
-  engineName: 'OIBus-main-thread',
+  engineName: 'OIBus-main',
   logParameters: {
     consoleLog: { level: 'debug' },
     fileLog: {
       level: 'debug',
-      fileName: './main-thread-journal.log',
+      fileName: `${path.parse(configFile).dir}/main-thread-journal.log`,
```
jfhenon, line 33: should the name be OIBus-main-journal.log?

src/server/controllers/logController.js
```diff
@@ -22,4 +22,52 @@ const getLogs = async (ctx) => {
...
+          case 'warn':
+            ctx.app.logger.warn(formattedLog)
+            break
+
+          case 'error':
+            ctx.app.logger.error(formattedLog)
+            break
+
+          default:
+            break
```
jfhenon, line 66: maybe a warning here?

jfhenon (commented):
2 comments

src/server/controllers/logController.js
burgerni10: So, I rather use console.warn to avoid spamming the logger, what do you think ?

PR #1652, opened 2022-03-07
feat(bulk): CRUD endpoints for HistoryQuery

kukukk:


src/HistoryQuery/HistoryQueryRepository.class.js
```diff
@@ -0,0 +1,162 @@
...
+   * @param {object} historyQuery - The HistoryQuery info
+   * @return {Promise<object>} - The HistoryQuery
+   */
+  async create(historyQuery) {
+    const query = `INSERT INTO ${HistoryQueryRepository.TABLE} (id, order, name, status, enabled, paused, southId, northId, startTime, endTime,
+                                                                filePattern, compress, settings)
+                   VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`
+    const stmt = await this.database.prepare(query)
+    await stmt.run(
+      historyQuery.uuid,
```
burgerni10, line 48: We should use id instead of uuid to be consistent with other entities of the project.

PR #1666, opened 2022-03-15
fix{ui): Frontend improvements

dezse97:


src/client/components/table/TableRows.jsx
```diff
@@ -56,6 +56,7 @@ const TableRows = ({
   handleDelete,
   handleDuplicate,
   handleOrder,
+  isHistoryQuery,
```
burgerni10, line 59: I'm not fond of this param, which is too specific to the historyQuery. Moreover, as discussed, there are some problems with deletion of other elements of Table in OIBus. I accept the review with that in mind, and I'll try some workaround to fix the deletion overall. In the meantime, continue with the history query new endpoints

burgerni10 (approved):

PR #1673, opened 2022-03-16
logs-frontend-improvement

burgerni10:


burgerni10:
@jfhenon : I've had some difficulties with a test on this PR, so I've commented it. Let me know if you have some hints on how to fix the commented test.

src/server/Server.class.js
```diff
@@ -89,8 +89,7 @@ class Server {
         if (err.status === 401) {
           ctx.status = 401
           ctx.set('WWW-Authenticate', 'Basic')
-          this.app.logger.warn('UnauthorizedError: Bad Combination Login/Password')
-          ctx.body = JSON.stringify(err)
+          this.app.logger.debug('UnauthorizedError: Bad Combination Login/Password')
```
jfhenon, line 92: only traced on debug? not warn?

jfhenon (approved):

src/server/Server.class.js
burgerni10: The warning raise questions to clients for a normal action that is explained in the frontend. Moreover the log also occurs each time the connection expires. In such a context, I don't think a warning is adapted, nor the info. Hence I passed it in debug

PR #1707, opened 2022-04-01
fix-status-front

burgerni10:


src/client/South/SouthStatus.jsx
```diff
@@ -6,10 +6,14 @@ import PointsButton from './PointsButton.jsx'
 import { ConfigContext } from '../context/ConfigContext.jsx'
 import Table from '../components/table/Table.jsx'
 
-function isIsoDate(str) {
-  if (!/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(str)) return false
-  const d = new Date(str)
-  return d.toISOString() === str
+function formatValue(value) {
```
jfhenon, line 9: function duplicated. should this be extracted in a common file?

jfhenon (approved):
one small comment

src/client/South/SouthStatus.jsx
burgerni10: Indeed, I've moved the function in the helpers/utils file

jfhenon (approved):

Issue #1411, opened 2021-10-04
await for Promise in OPCHDA -> historyQuery()


kukukk:
Instead of the while loop in  OPCHDA -> historyQuery() await for a Promise which is resolved in handleResponse.
Example: https://stackoverflow.com/questions/26150232/resolve-javascript-promise-outside-function-scope/36072263#36072263

PR #1728, opened 2022-04-07
fix(engine): reset active south, north and history on engine reload

burgerni10:


src/HistoryQuery/HistoryQuery.class.js
```diff
@@ -121,7 +121,14 @@ class HistoryQuery {
...
       await this.south.init()
       await this.south.connect()
-      await this.export()
+      if (this.south.connected) {
+        await this.export()
+      } else {
+        await this.disable()
+        this.logger.error(`Could not connect to south connector ${
+          this.south.dataSource.name
+        }. This history query has been disable. Check the connection before enabling it again.`)
```
jfhenon, line 130: disabled

src/HistoryQuery/HistoryQuery.class.js
```diff
@@ -148,7 +155,14 @@ class HistoryQuery {
...
       await this.north.init()
       await this.north.connect()
-      await this.import()
+      if (this.north.connected) {
+        await this.import()
+      } else {
+        await this.disable()
+        this.logger.error(`Could not connect to north connector ${
+          this.north.application.name
+        }. This history query has been disable. Check the connection before enabling it again.`)
```
jfhenon, line 164: disabled

src/HistoryQuery/HistoryQuery.class.js
```diff
@@ -179,15 +193,21 @@ class HistoryQuery {
...
           this.logger.error(`scanMode ${scanMode} ignored: scanGroup.points undefined or empty`)
         }
       }
       this.statusData.scanGroup = null
       this.updateStatusDataStream()
     } else {
-      await this.exportScanMode(this.dataSource.scanMode)
+      const exportResult = await this.exportScanMode(this.dataSource.scanMode)
+      if (exportResult === -1) {
+        return
```
jfhenon, line 209: should we trace the condition?

src/HistoryQuery/HistoryQuery.class.js
```diff
@@ -179,15 +193,21 @@ class HistoryQuery {
           this.statusData.scanGroup = `${scanGroup.scanMode} - ${scanGroup.aggregate}`
           this.updateStatusDataStream()
           // eslint-disable-next-line no-await-in-loop
-          await this.exportScanMode(scanMode)
+          const exportResult = await this.exportScanMode(scanMode)
+          if (exportResult === -1) {
+            return
```
jfhenon, line 198: should we trace the condition?

src/HistoryQuery/HistoryQuery.class.js
```diff
@@ -271,7 +291,13 @@ class HistoryQuery {
       }
 
       // eslint-disable-next-line no-await-in-loop
-      await this.south.historyQuery(scanMode, startTime, intervalEndTime)
+      const historyQueryResult = await this.south.historyQuery(scanMode, startTime, intervalEndTime)
+
+      if (historyQueryResult === -1) {
+        this.logger.error(`Error while retrieving data. Exiting historyQueryHandler. startTime: ${
+          startTime.toISOString()}, intervalEndTime: ${intervalEndTime.toISOString()}`)
+        return -1
```
jfhenon, line 299: return -1 or 0 ? should we return true/false instead?

src/north/console/Console.class.js
```diff
@@ -17,6 +17,7 @@ class Console extends ApiHandler {
     this.canHandleValues = true
     this.canHandleFiles = true
     this.verbose = applicationParameters.Console.verbose ?? false
+    this.connected = true
```
jfhenon, line 20: `this.connected = true` should it be instead set if the connect() has been successful in the parent class?

src/north/csvtohttp/CsvToHttp.class.js
```diff
@@ -46,6 +46,7 @@ class CsvToHttp extends ApiHandler {
 
     this.canHandleFiles = true
     this.canHandleValues = false
+    this.connected = true
```
jfhenon, line 49: `this.connected = true` should it be instead set if the connect() has been successful in the parent class?

src/north/filewriter/FileWriter.class.js
```diff
@@ -21,6 +21,7 @@ class FileWriter extends ApiHandler {
     this.outputFolder = path.resolve(applicationParameters.FileWriter.outputFolder)
     this.prefixFileName = applicationParameters.FileWriter.prefixFileName ?? ''
     this.suffixFileName = applicationParameters.FileWriter.suffixFileName ?? ''
+    this.connected = true
```
jfhenon, line 24: `this.connected = true` should it be instead set if the connect() has been successful in the parent class?

src/south/OPCHDA/OPCHDA.class.js
```diff
@@ -69,6 +69,7 @@ class OPCHDA extends ProtocolHandler {
       // eslint-disable-next-line no-await-in-loop
       await this.delay(100)
     }
+    return 0
```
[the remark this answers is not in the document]
burgerni10: I agree, we can do that with issue #1411

src/HistoryQuery/HistoryQuery.class.js
burgerni10: I didn't because normally it is trace in the exportScanMode method. I can add a logger.trace here though

src/HistoryQuery/HistoryQuery.class.js
burgerni10: I initially planned to add other error codes, but ended with only two. I prefer specifying -1 since some method returns the number of values so there won't be any chance to mistake a return value true/false with a number (1/0)

src/north/console/Console.class.js
burgerni10: I disagree: the parent class is called first to initiate some variables. For some connectors, like MQTT (both for North and South) the connector is connected in async method. So we cannot mark it connected too soon. I know it feels like redondant, but I didn't see another way to manage this case.

jfhenon: see branch https://github.com/OptimistikSAS/OIBus/tree/consistent-connect

burgerni10: Done, thank you !

jfhenon (approved):

PR #1774, opened 2022-05-11
fix-rest-api

burgerni10:


src/south/OPCUA_HA/OPCUA_HA.class.js
```diff
@@ -199,10 +199,21 @@ class OPCUA_HA extends ProtocolHandler {
     Object.keys(logs).forEach((statusCode) => {
       switch (statusCode) {
         case StatusCodes.BadIndexRangeNoData: // No data exists for the requested time range or event filter.
-          this.logger.debug(`${logs[statusCode].description} (${statusCode}): ${logs[statusCode].affectedNodes.toString()}`)
+          if (logs[statusCode].affectedNodes.length > 10) {
```
kukukk, line 202: Where this `10` is coming from? It would be better to define a constant for it with a descriptive name.

src/south/OPCUA_HA/OPCUA_HA.class.js
```diff
@@ -199,10 +199,21 @@ class OPCUA_HA extends ProtocolHandler {
...
+            this.logger.debug(`${logs[statusCode].description} (${statusCode}): [${
+              logs[statusCode].affectedNodes[0]}..${logs[statusCode].affectedNodes[logs[statusCode].affectedNodes.length - 1]}]`)
+          } else {
+            this.logger.debug(`${logs[statusCode].description} (${statusCode}): [${logs[statusCode].affectedNodes.toString()}]`)
+          }
+
           break
         default:
-          this.logger.debug(`${logs[statusCode].description} (${statusCode}): ${logs[statusCode].affectedNodes.toString()}`)
+          if (logs[statusCode].affectedNodes.length > 10) {
```
kukukk, line 211: I can't see any difference compared to the previous case.

PR #1788, opened 2022-05-20
fix(opcua): manage certificate generation with selfsigned package instead of downloading openssl at runtime

burgerni10:


src/south/OPCUA_DA/OPCUA_DA.class.js
```diff
@@ -57,6 +66,37 @@ class OPCUA_DA extends ProtocolHandler {
     await this.connectToOpcuaServer()
   }
 
+  async init() {
+    await super.init()
+    await this.initOpcuaCertificateManager()
+  }
+
+  async initOpcuaCertificateManager() {
```
kukukk, line 74: Is it the same method as in OPCUA_HA? Can't we refactor it into the EncryptionService or a new service to prevent code duplication?

PR #1891, opened 2022-08-12
status-data

burgerni10:
This PR start to remove statusData from connectors to be managed in Protocol / Api Handlers. 
I'll go further when PR #1797 will be done


src/north/MongoDB/MongoDB.class.js
```diff
@@ -56,10 +56,11 @@ class MongoDB extends ApiHandler {
     this.logger.trace(`MongoDB handleValues() call with ${values.length} values`)
     try {
       await this.makeRequest(values)
-      this.statusData['Last handled values at'] = new Date().toISOString()
-      this.statusData['Number of values sent since OIBus has started'] += values.length
-      this.statusData['Last added point id (value)'] = `${values[values.length - 1].pointId} (${JSON.stringify(values[values.length - 1].data)})`
-      this.updateStatusDataStream()
+      this.updateStatusDataStream({
```
jfhenon, line 59: Most handleValues return the same information should the sse call be  in the parent class ?

jfhenon (commented):
See comment

src/north/MongoDB/MongoDB.class.js
burgerni10: I agree, but I don't think it should be done here, rather in the cache refactor in PR #1797

src/north/CsvToHttp/CsvToHttp.class.js
```diff
@@ -123,7 +123,11 @@ class CsvToHttp extends ApiHandler {
           this.statusData['Last Parsed file'] = filePath
```
jfhenon, line 123: should be removed

jfhenon (changes requested):

PR #1912, opened 2022-08-19
sqlite-deps

burgerni10:
Only use better-sqlite3
Remove sqlite3 and sqlite dependencies

better-sqlite3 offers better performance for writing and retrieving values, and works in worker-thread (used in pino), hence this choice.


marouanehassanioptimistik (changes requested):
i think we should write some kind of testing for the database.service.js to ensure we are not having regressions even if the rewrite looks pretty straightforward

burgerni10:
I've added tests for each method, and remove the await / async since better-sqlite3 is sync

marouanehassanioptimistik (approved):

PR #1918, opened 2022-08-24
fix(opcua): fix certificate management and reconnection ; switch to node-opcua-client (lighter)

burgerni10:
We only need session, no client + session. The connect and disconnect functions are cleaner.

Also, since OIBus is client only, I use node-opcua-client, removing all node-opcua server dependencies.

I needed to add an "autotrust" certificate to true, otherwise, it was failing to connect to OPCUA servers.


src/south/OPCUA/OPCUA_DA/OPCUA_DA.class.js
```diff
@@ -147,7 +156,7 @@ class OPCUA_DA extends ProtocolHandler {
       return
     }
 
-    this.subscription = Opcua.ClientSubscription.create(this.session, {
+    this.subscription = ClientSubscription.create(this.session, {
```
marouanehassanioptimistik, line 159: i am not completely sure why this code is commented and why we would make change to commented code. Should we plainly remove this piece of code?

marouanehassanioptimistik (approved):
ok for the changes just a remark about commented code

src/south/OPCUA/OPCUA_DA/OPCUA_DA.class.js
burgerni10: This subscription was developed by Szabi a while ago, but the management of subscription associated to the points needed more conception. So this piece of code has been commented for future used.

marouanehassanioptimistik: i find that usually very unpractical to leave commented code, we should open a branch with this rather than cluttering code with comments (and we don't even know why this is commented)

PR #1937, opened 2022-08-30
sqlite-fix

burgerni10:


jfhenon (commented):
Is this possible that startTime or endTime are not in the Date format? (as we don't catch or test the result of getTime()

burgerni10:
> Is this possible that startTime or endTime are not in the Date format? (as we don't catch or test the result of getTime()

No, startTime and endTime always come from SouthConnector historyQuery which passes Date arguments.

Typescript will make sure of that at the build time :)

burgerni10:
> > Is this possible that startTime or endTime are not in the Date format? (as we don't catch or test the result of getTime()
> 
> No, startTime and endTime always come from SouthConnector historyQuery which passes Date arguments.
> 
> Typescript will make sure of that at the build time :)

Moreover, I started [to refactor the connectors](https://github.com/OptimistikSAS/OIBus/pull/1921) and the engine to make sure the connectors do not catch any errors, but let them throw it to the main class (South/North connector).

PR #1958, opened 2022-09-13
fix(ui): 1807 solved button display problem when renaming connector

Reloflex:


burgerni10 (changes requested):
image
The sub nav grows significantly in Edit mode. Can you look at it ?

Also, I find your feedback more explicit than the icon. Can you remove the icons in edit and read modes ?

src/client/North/Form/NorthForm.jsx
```diff
@@ -58,16 +58,19 @@ const NorthForm = ({ application, applicationIndex, onChange }) => {
...
-              outline
-              onClick={() => {
-                setRenamingConnector(`north-${application.id}`)
-                setPencil(false)
-              }}
-              className="util-button"
-            >
-              <FaPencilAlt className="oi-icon ms-2" />
-            </Button>
+          <Button
```
burgerni10, line 61: These changes should also be reflected in the SouthForm.jsx file

src/client/components/EditableIdField.jsx
```diff
@@ -46,9 +46,18 @@ const EditableIdField = ({
         name={connectorName}
         inline
       />
-      <Button outline onClick={handleDoneEditName} className="util-button">
-        <FaCheck className="oi-icon ms-2" />
-      </Button>
+      <h6 className="align-items-top mb-3">
```
burgerni10, line 49: The mb-3 css class is probably the cause of the sub-nav growth in edit mode

src/client/North/Form/NorthForm.jsx
Reloflex: Yes, I forgot to add the changes to the other instances of the same edit. Will do

src/client/components/EditableIdField.jsx
Reloflex: The sub nav grows in edit mode to give space for the potential Error message below the edit. I kept the space for it and aligned the button after the edit to the same level.

burgerni10: Would it be possible to add this space only if the error occurs ? It feels weird to have it appear as soon as we set the edit mode and no error appear.

burgerni10 (approved):

PR #1971, opened 2022-09-23
log-folder

burgerni10:


src/migration/migrationRules.js
```diff
@@ -1463,5 +1463,44 @@ module.exports = {
...
+      await fs.rename(oldLogSqlFileName, newFileLogSqlPath)
+    } catch (err) {
+      if (err.code !== 'ENOENT') {
+        logger.error(err)
+      }
+    }
+    delete config.engine.logParameters.sqliteLog.fileName
+
+    try {
+      await fs.rm(path.resolve('./migration-journal.log'), { force: true })
```
kukukk, line 1492: Is it ok to just delete the old migration log?

src/migration/migrationRules.js
```diff
@@ -1463,5 +1463,44 @@ module.exports = {
...
+    } catch (err) {
+      if (err.code !== 'ENOENT') {
+        logger.error(err)
+      }
+    }
+    try {
+      await fs.rm(path.resolve('./OIBus-main-journal.log'), { force: true })
```
kukukk, line 1499: Is it ok to just delete the old main journal log?

src/migration/migrationRules.js
burgerni10: My point is that both these files (migration-journal.log and OIBus-main-journal.log) are used for debugging purposes when OIBus fails to (re)start. I think it's safe to remove them now.

@jfhenon what is your opinion on that ?

PR #1997, opened 2022-09-27
fix(cache): clean up and move cache files

burgerni10:
Certs folder is removed to be placed in the documentation


src/HistoryQuery/HistoryQueryEngine.class.js
```diff
@@ -47,7 +47,7 @@ class HistoryQueryEngine extends BaseEngine {
       this.logger.debug(`Creating history cache folder "${this.cacheFolder}".`)
       await fs.mkdir(this.cacheFolder, { recursive: true })
     }
-    this.historyQueryRepository = new HistoryQueryRepository(path.resolve(CACHE_FOLDER, 'history-query.db'))
+    this.historyQueryRepository = new HistoryQueryRepository(path.resolve('./history-query.db'))
```
kukukk, line 50: Would not be better to create some constants instead of hard-coding? Also for other file names, cache folder and sub-folders.

burgerni10: I totally agree. I will update the PR. Does it seem ok otherwise ?

kukukk: Yep.

Issue #1809, opened 2022-06-20
[Logs] - Paginate logs


burgerni10:
On the backend, make page requests to the db to lower the payload

When retrieving logs, use a searchCriteria having  : 
-startDate
-endDate
-verbosity
-contains
-scope
-page
-pageSize

The frontend must be improved to reflect this searchCriteria, and come close to our other interfaces, similar to something like this : 
![image](https://user-images.githubusercontent.com/28055217/174586868-e4cfa74c-aea7-410b-8258-a082ce807ebd.png)

The pagination component should be similar to this : 
![image](https://user-images.githubusercontent.com/28055217/174586975-18e8a7a2-db70-4103-90c6-cef0d800fcf6.png)

PR #2000, opened 2022-09-28
paginated-logs-backend

burgerni10:
Prepare paginate logs #1809 on the backend

The filtering for text content is now possible on the backend. It was previously done on the client side with all logs.

Still to implement by @Reloflex: 
- A pagination component to navigate between page
- A form as shown in #1809
- A rework of the displayed logs
- A hint that now logs match the query if no log is retrieved

### DIFF
--- a/src/services/database.class.spec.js
+++ b/src/services/database.class.spec.js
@@ -258,6 +258,90 @@ describe('Database services', () => {
     expect(result).toEqual(retrievedLogs)
   })
 
+  it('should get paginated logs', () => {
+    const retrievedLogs = [
+      { timestamp: 123, message: 'myLogMessage1' },
+      { timestamp: 124, message: 'myLogMessage2' },
+    ]
+    const localAll = jest.fn()
+    localAll.mockReturnValue(retrievedLogs)
+    const localGet = jest.fn()
+    localGet.mockReturnValue({ count: 2 })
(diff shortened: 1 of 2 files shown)


jfhenon (commented):
do we still need getLogs() and, if yes, should this call getPaginatedLogs behind the scene?

burgerni10:
> do we still need getLogs() and, if yes, should this call getPaginatedLogs behind the scene?

Yes getLogs is still used. We cannot remove it without refactoring the whole log page (form, results querying). I suggest to remove it once the rest of the work will be done.

marouanehassanioptimistik (changes requested):
i think the sorting should be passed as a parameter. also it is more logic to sort logs by timestamp descending by default

burgerni10:
> i think the sorting should be passed as a parameter. also it is more logic to sort logs by timestamp descending by default



> i think the sorting should be passed as a parameter. also it is more logic to sort logs by timestamp descending by default

My mistake, it should indeed be descending.

However, I don't think it makes sense to have a sorting parameter: I don't see a use case where we would like to go through the logs from the start.

marouanehassanioptimistik:
it seems best practice to me for a database service not to suppose what the use is for it is at least more flexible. We don't necessarily need to allow another sorting at ui level and it is up to the rest controller to hard code the sorting. Just à thought you can hardcode the sorting in the service if you judge it is overkill design.

PR #2028, opened 2022-10-21
north-fix

burgerni10:
In some cases, a file is being written by a south when the north scan its cache folder, yielding an error that was not caught.

Also, since the cache database does not exist anymore for files, the following block is not useful anymore, because  the retrieveFileFromCache method already returns the oldest file in cache.

```js
try {
      await fs.stat(fileToSend.path)
    } catch (error) {
```

### DIFF
--- a/src/engine/cache/file-cache.js
+++ b/src/engine/cache/file-cache.js
@@ -111,17 +111,34 @@ class FileCache extends BaseCache {
    * @returns {Promise<{path: string, timestamp: number}|null>} - The file to send
    */
   async retrieveFileFromCache() {
-    const fileNames = await fs.readdir(this.fileFolder)
+    let fileNames = []
+    try {
+      fileNames = await fs.readdir(this.fileFolder)
+    } catch (error) {
+      this.logger.error(error)
+    }
 
     if (fileNames.length === 0) {
       return null
     }
 
     const sortedFiles = fileNames
-      .map(async (fileName) => ({
-        path: path.resolve(this.fileFolder, fileName),
-        timestamp: (await fs.stat(path.resolve(this.fileFolder, fileName))).mtime.getTime(),
-      }))
+      .map(async (fileName) => {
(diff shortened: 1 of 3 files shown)


src/engine/cache/file-cache.js
```diff
@@ -118,10 +118,19 @@ class FileCache extends BaseCache {
...
+          fileState = await fs.stat(path.resolve(this.fileFolder, fileName))
+        } catch (error) {
+          this.logger.error(error)
+        }
+        return {
+          path: path.resolve(this.fileFolder, fileName),
+          timestamp: fileState ? fileState.mtime.getTime() : null,
+        }
+      })
+      .filter((file) => file.timestamp !== null)
```
marouanehassanioptimistik, line 133: adding comment here would be useful since it is not particularly trivial what is achieved

src/north/north-connector.js
```diff
@@ -243,23 +242,20 @@ class NorthConnector {
...
+        this.resetFilesTimeout(this.caching.sendInterval)
+        return
+      }
     } catch (error) {
-      // File in cache does not exist on filesystem
-      await this.fileCache.removeFileFromCache(fileToSend.path, false)
-      this.logger.error(`File "${fileToSend.path}" not found! The file has been removed from the cache.`)
-      this.resetFilesTimeout(this.caching.sendInterval)
+      // Error triggered when a file is being written (from a South) during the retrieveFileFromCache call
+      this.logger.error(error)
```
marouanehassanioptimistik, line 255: is this catch block useful can an error be thrown before?

burgerni10: fs.readdir can throw an error in the retrieveFileFromCache method if the folder does not exist. I'm gonna remove this logger.error and place the try catch in retrieveFileFromCache method around readdir

src/engine/cache/file-cache.js
burgerni10: Indeed, on it.

PR #2035, opened 2022-10-24
fix(north): fix file cache stat eperm error

burgerni10:
A fix still need to be done on the way the request service (axios or fetch) manage the readStream that can lock a file in a cache while OIBus is running, but this PR fix the lock on one specific file when the north read the cache directory

### DIFF
--- a/src/engine/cache/file-cache.js
+++ b/src/engine/cache/file-cache.js
@@ -54,18 +54,28 @@ class FileCache extends BaseCache {
     await createFolder(this.fileFolder)
     await createFolder(this.errorFolder)
 
-    const files = await fs.readdir(this.fileFolder)
-    if (files.length > 0) {
-      this.logger.debug(`${files.length} files in cache.`)
-    } else {
-      this.logger.debug('No files in cache.')
+    try {
+      const files = await fs.readdir(this.fileFolder)
+      if (files.length > 0) {
+        this.logger.debug(`${files.length} files in cache.`)
+      } else {
+        this.logger.debug('No files in cache.')
+      }
+    } catch (error) {
+      // If the folder does not exist, an error is logged but OIBus keep going to check errors and archive folders
+      this.logger.error(error)
     }
 
-    const errorFiles = await fs.readdir(this.errorFolder)
-    if (errorFiles.length > 0) {
-      this.logger.warn(`${errorFiles.length} files in error cache.`)
-    } else {
-      this.logger.debug('No error files in cache.')
+    try {
+      const errorFiles = await fs.readdir(this.errorFolder)
+      if (errorFiles.length > 0) {
+        this.logger.warn(`${errorFiles.length} files in error cache.`)
+      } else {
+        this.logger.debug('No error files in cache.')
+      }
+    } catch (error) {
+      // If the folder does not exist, an error is logged but not thrown if the file cache folder is accessible
+      this.logger.error(error)
     }
 
     if (this.archiveFiles) {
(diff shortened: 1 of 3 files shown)


src/engine/cache/file-cache.js
```diff
@@ -54,18 +54,26 @@ class FileCache extends BaseCache {
...
-      this.logger.debug('No files in cache.')
+    try {
+      const files = await fs.readdir(this.fileFolder)
+      if (files.length > 0) {
+        this.logger.debug(`${files.length} files in cache.`)
+      } else {
+        this.logger.debug('No files in cache.')
+      }
+    } catch (error) {
+      this.logger.error(error)
```
marouanehassanioptimistik, line 65: a comment explaining why this can happen and why silently ignoring it is the right way to go would be more than helpful

src/engine/cache/file-cache.js
```diff
@@ -54,18 +54,26 @@ class FileCache extends BaseCache {
...
-      this.logger.debug('No error files in cache.')
+    try {
+      const errorFiles = await fs.readdir(this.errorFolder)
+      if (errorFiles.length > 0) {
+        this.logger.warn(`${errorFiles.length} files in error cache.`)
+      } else {
+        this.logger.debug('No error files in cache.')
+      }
+    } catch (error) {
+      this.logger.error(error)
```
marouanehassanioptimistik, line 76: a comment explaining why this can happen and why silently ignoring it is the right way to go would be more than helpful

src/engine/cache/file-cache.js
```diff
@@ -122,24 +130,22 @@ class FileCache extends BaseCache {
...
         try {
-          // Error triggered when a file is being written (from a South)
-          fileState = await fs.stat(path.resolve(this.fileFolder, fileName))
+          const stat = await fs.stat(path.resolve(this.fileFolder, fileName))
+          fileStats.push({
+            path: path.resolve(this.fileFolder, fileName),
+            timestamp: stat.mtime.getTime(),
+          })
         } catch (error) {
           this.logger.error(error)
```
marouanehassanioptimistik, line 143: a comment explaining why this can happen and why silently ignoring it is the right way to go would be more than helpful

src/engine/cache/file-cache.js
```diff
@@ -195,7 +201,12 @@ class FileCache extends BaseCache {
    * @returns {Boolean} - Cache empty or not
    */
   async isEmpty() {
-    const files = await fs.readdir(this.fileFolder)
+    let files = []
+    try {
+      files = await fs.readdir(this.fileFolder)
+    } catch (error) {
+      this.logger.error(error)
```
marouanehassanioptimistik, line 208: a comment explaining why this can happen and why silently ignoring it is the right way to go would be more than helpful

src/engine/cache/file-cache.js
```diff
@@ -210,9 +221,15 @@ class FileCache extends BaseCache {
       clearTimeout(this.archiveTimeout)
     }
 
-    const files = await fs.readdir(this.archiveFolder)
-    const referenceDate = new Date().getTime()
+    let files = []
+    try {
+      files = await fs.readdir(this.fileFolder)
+    } catch (error) {
+      this.logger.error(error)
```
marouanehassanioptimistik, line 228: a comment explaining why this can happen and why silently ignoring it is the right way to go would be more than helpful

src/engine/cache/file-cache.js
```diff
@@ -231,8 +248,13 @@ class FileCache extends BaseCache {
    * @returns {Promise<void>} - The result promise
    */
   async removeFileIfTooOld(filePath, referenceDate, archiveFolder) {
-    const stats = await fs.stat(path.join(archiveFolder, filePath))
-    if (stats.mtimeMs + this.retentionDuration < referenceDate) {
+    let stats
+    try {
+      stats = await fs.stat(path.join(archiveFolder, filePath))
+    } catch (error) {
+      this.logger.error(error)
```
marouanehassanioptimistik, line 255: a comment explaining why this can happen and why silently ignoring it is the right way to go would be more than helpful

marouanehassanioptimistik (changes requested):

PR #2052, opened 2022-10-28
story: remove sqlite for cache

burgerni10:
This is the first step of improving the value cache:
- I firstly removed the engine dependency from the south/north connectors
- I removed the request service (keeping only node-fetch) and used an utils function
- I had to change de authentication logic to reuse the code as an util function
- I had to change the proxy logic a bit to reuse the code as an util function
- Adapt north/south and engine with these changes
- Change some settings (HTTP, configuration) now specific to each North instead of the engine

With these changes, the next steps are:
- Moving the south buffer logic into each north (flush with max/timeout)
- Write values to North cache in a file and not in a database
- Rethink the logic of groupCount / maxGroupCount to adapt the sharding with these parameters

This branch is aimed to be the main branch for the future discussed changes


src/service/utils.js
```diff
@@ -2,29 +2,130 @@ const fs = require('node:fs/promises')
 const { createReadStream, createWriteStream } = require('node:fs')
```
marouanehassanioptimistik, line 2: it is probably my java deformation but rather than naming this utils i would name `http-request-static-functions` or if you really like utils `http-request-utils-functions`

src/engine/health-signal.js
```diff
@@ -21,7 +23,7 @@ class HealthSignal {
     this.http.proxy = Array.isArray(engineConfig.proxies) ? engineConfig.proxies.find(({ name }) => name === this.http.proxy) : null
     this.httpTimer = null
     this.loggingTimer = null
-    this.engineName = engineConfig.engineName
+    this.oibusName = engineConfig.engineName
```
marouanehassanioptimistik, line 26: haven't u renamed as well in the engine config?

src/north/north-connector.js
```diff
@@ -44,14 +47,12 @@ class NorthConnector {
     this.logParameters = configuration.logParameters
     this.caching = configuration.caching
     this.subscribedTo = configuration.subscribedTo
+    this.proxies = proxies
```
marouanehassanioptimistik, line 50: can't we simplify a bit this, we should provide an array of proxies and read from config the selected proxy and directly create a proxyAgent as sole internal variable. Does this make sense?

src/south/south-ads/south-ads.js
```diff
@@ -12,16 +12,26 @@ class SouthADS extends SouthConnector {
...
-    super(configuration, engine, {
-      supportListen: false,
-      supportLastPoint: true,
-      supportFile: false,
-      supportHistory: false,
-    })
+  constructor(
+    configuration,
+    engineAddValues,
+    engineAddFiles,
```
marouanehassanioptimistik, line 22: engineAddFilesCallback?

src/south/south-connector.js
```diff
@@ -34,13 +37,21 @@ class SouthConnector {
...
+   * @param {Function} engineAddValues - The Engine add values method
+   * @param {Function} engineAddFiles - The Engine add file method
    * @param {Object} supportedModes - The supported modes
    * @return {void}
    */
-  constructor(configuration, engine, supportedModes = {}) {
+  constructor(
+    configuration,
+    engineAddValues,
+    engineAddFiles,
```
marouanehassanioptimistik, line 48: same remark here regarding naming

src/web-server/web-server.js
```diff
@@ -49,6 +49,10 @@ class Server {
     this.logger.setEncryptionService(this.encryptionService)
   }
 
+  /**
+   * Start the web server
+   * @returns {Promise<void>} - The result promise
```
marouanehassanioptimistik, line 54: i like this doc :)

marouanehassanioptimistik (changes requested):
i like the overall look of this, i made mainly naming comments and other design that may or may not be applicable let me know

src/north/north-connector.js
burgerni10: I saw several "problems" with that :
- the proxy to use is set in the connector specific settings, not in the global north settings, so I prefer not to retrieve it from the main north constructor as it is now. That is why the proxySettings are set in AWS3, OIA and OIC connectors, not in NorthConnector constructor
- The ProxyAgent constructor is async and must be called in the start method. One possibility was to add the proxy list in the start method arguments, but it is used by only three connectors.

Overall, I suggest to make a separate dev issue for that.

src/north/north-influx-db/north-influx-db.js
```diff
@@ -170,7 +177,14 @@ class NorthInfluxDB extends NorthConnector {
     })
 
     const headers = { 'Content-Type': 'application/x-www-form-urlencoded' }
-    await this.engine.requestService.httpSend(url, 'POST', null, null, body, headers)
+    await httpSend(
+      url,
+      'POST',
+      headers,
+      body,
+      this.caching.timeout,
```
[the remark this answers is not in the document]
burgerni10: It's the same timeout than originally used before the PR, but I agree it should be a specific timeout associated to the north connector. I'll make an issue for that to manage it in a separate PR.

marouanehassanioptimistik:
@burgerni10 is there anything blocking the merge of this?

burgerni10:
> @burgerni10 is there anything blocking the merge of this?

I use this branch as a story branch for the other PR to test all of the changes. I rename the PR so it is clearer

marouanehassanioptimistik:
ok

PR #2055, opened 2022-10-31
fix(opcua): Prevent duplicate session creation

kukukk:


burgerni10 (changes requested):
I think a `this.disconnecting = false` is missing (in both connectors) at the end of the internalDisconnect Method

Otherwise, it will be lock to true on config reload (for example)

Also, can you add a test case (in both connectors) to check that the method is not called on error while disconnecting is already true ?

kukukk:
During the config reload the engine is stopped, then started, which should create a new south instance where `disconnecting` will be `false`.
I will add the test cases.

burgerni10 (approved):

PR #2063, opened 2022-11-07
fix(cache): refactor values cache

burgerni10:
Unit tests must be implemented. Just want to check the logic for now.


burgerni10:
I eventually used a deferred promise in the value cache. Next steps are: 
- run cache if support values/files
- migrate values cache
- create a separate archive service (for now in the file cache logic)
- same logic for file cache

src/service/cache/value-cache.service.js
```diff
@@ -0,0 +1,447 @@
...
+const RESEND_IMMEDIATELY_TIMEOUT = 100
+
+const VALUE_FOLDER = 'values'
+const ERROR_FOLDER = 'values-errors'
+
+const BUFFER_FILE_NAME = 'buffer.tmp'
+
+/**
+ * Local cache implementation to group events and store them when the communication with the North is down.
+ */
```
marouanehassanioptimistik, line 20: we should add a readme.md explaining the design of the cache and refer to it. Please make an issue for this

src/service/cache/value-cache.service.js
```diff
@@ -0,0 +1,447 @@
...
+    this.sendingValues$ = null
+
+    this.valueFolder = path.resolve(this.baseFolder, VALUE_FOLDER)
+    this.errorFolder = path.resolve(this.baseFolder, ERROR_FOLDER)
+  }
+
+  /**
+   * Create databases and folders
+   * @returns {Promise<void>} - The result promise
+   */
```
marouanehassanioptimistik, line 64: a synthetic reminder of what is done at startup and a reference to the readme would be quite helpful

src/service/cache/value-cache.service.js
```diff
@@ -0,0 +1,447 @@
...
+    } else {
+      this.logger.warn('No send interval. No values will be sent.')
+    }
+  }
+
+  /**
+   * Method used to flush the buffer from a time trigger or a max trigger
+   * @param {'time-flush' | 'max-flush'} flag - The trigger
+   * @returns {Promise<void>} - The result promise
+   */
```
marouanehassanioptimistik, line 137: Explain synthetically what a flush consist in

marouanehassanioptimistik (approved):
i made comment only on documentation as i find it light overall the code looks good to me. You can merge and improve the doc later but we really should do it.

PR #2064, opened 2022-11-07
fix(cache): Remove cache folder when removing connector

kukukk:


src/service/configuration.service.js
```diff
@@ -105,8 +105,27 @@ class ConfigurationService {
...
+    const folders = await fs.readdir(dataStreamFolderPath)
+
+    // eslint-disable-next-line no-restricted-syntax
+    for (const folder of folders) {
+      const uid = folder.replace('north-', '').replace('south-', '')
+      if (!idList.includes(uid)) {
+        // eslint-disable-next-line no-await-in-loop
+        await fs.rmdir(path.resolve(dataStreamFolderPath, folder), { recursive: true })
+      }
+    }
```
burgerni10, line 127: Can you use a promise syntax instead of this for loop, since the eslint rules is set to avoid that. 

See the following snippet for example
https://github.com/OptimistikSAS/OIBus/blob/fb8c0ed0a8391375056ff857cc46c7df4d55a00c/src/engine/cache/file-cache.js#L137-L151

burgerni10 (changes requested):

burgerni10 (approved):

PR #2074, opened 2022-11-14
fix(cache): migrate values cache and file errors folder

burgerni10:
The migration of a 8,19GB outputs 6GB of files (6000 files). My guess is that the database stores uri encoded values where the new cache store regular JSON, reducing the size of the string stored. Such a migration takes 60s approximately on Mac M1.

With many files (~6000), OIBus initialisation takes some time (5/10s on Mac) at full CPU usage because the files are parsed to count the number of values.

### DIFF
--- a/src/migration/migration-rules.js
+++ b/src/migration/migration-rules.js
@@ -6,7 +6,7 @@ const db = require('better-sqlite3')
 const { nanoid } = require('nanoid')
 const databaseMigrationService = require('./database-migration.service')
 const databaseService = require('../service/database.service')
-const { createFolder } = require('../service/utils')
+const { createFolder, filesExists } = require('../service/utils')
 
 module.exports = {
   2: (config, logger) => {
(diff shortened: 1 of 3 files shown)


src/migration/migration-rules.js
```diff
@@ -1605,6 +1605,80 @@ module.exports = {
...
+
+        try {
+          await fs.rm(path.resolve(`./cache/data-stream/north-${north.id}/values.db`), { force: true })
+          await fs.rm(path.resolve(`./cache/data-stream/north-${north.id}/values-error.db`), { force: true })
+        } catch (err) {
+          if (err.code !== 'ENOENT') {
+            logger.error(err)
+          }
+        }
+      }
```
marouanehassanioptimistik, line 1681: should we delete the file only if the migration succeeded? with all try catch i am not sure what is done finally

burgerni10: If the database is migrated, then the database for this specific north can be removed. In case of removal error I think the migration should go on with other connectors instead of blocking. I add a code comment to explain that

PR #2105, opened 2022-11-30
story: ESM with Typescript

burgerni10:


src/engine/base-engine.js
```diff
@@ -69,6 +93,10 @@ class BaseEngine {
    * @returns {Promise<void>} - The result promise
    */
   async initEngineServices(engineConfig) {
+    // const packageJson = JSON.parse(await fs.readFile('package.json'))
+    // this.version = packageJson.version
+    // TODO
+    this.version = '2.4.0'
```
marouanehassanioptimistik, line 99: is this intended to be left and fixed in another PR?

src/index.js
```diff
@@ -44,14 +42,17 @@ if (cluster.isMaster) {
   const baseDir = path.resolve(path.extname(configFile) ? path.parse(configFile).dir : configFile)
   process.chdir(baseDir)
   createFolder(baseDir).then(async () => {
+    // TODO
+    const packageJson = '2.4.0' // JSON.parse(await fs.readFile('package.json'))
```
marouanehassanioptimistik, line 46: same here

marouanehassanioptimistik (changes requested):
the overall look seems ok i have seen some weird todos leftover are they fixed later?

src/engine/base-engine.js
burgerni10: Fixed in another PR

Issue #2065, opened 2022-11-08
[Dev Issue] - Proxy settings management


burgerni10:
can't we simplify a bit this, we should provide an array of proxies and read from config the selected proxy and directly create a proxyAgent as sole internal variable. Does this make sense?

_Originally posted by @marouanehassanioptimistik in https://github.com/OptimistikSAS/OIBus/pull/2052#discussion_r1015905190_

PR #2120, opened 2022-12-06
feat(proxy): use a proxy service to manage proxy in connectors

burgerni10:
Fix issue #2065 

Proxy service is now available from south connector too, since it can be used for retrieving data from REST API.


marouanehassanioptimistik (approved):
LGTM

PR #2132, opened 2022-12-17
story-angular-engine-repository

burgerni10:


backend/repository/engine.repository.ts
```diff
@@ -0,0 +1,264 @@
...
+      `health_signal_http_verbose AS healthSignalHttpVerbose, ` +
+      `health_signal_http_address AS healthSignalHttpAddress, ` +
+      `health_signal_http_proxy_id AS healthSignalHttpProxyId, ` +
+      `health_signal_http_authentication_type AS healthSignalHttpAuthenticationType, ` +
+      `health_signal_http_authentication_key AS healthSignalHttpAuthenticationKey, ` +
+      `health_signal_http_authentication_secret AS healthSignalHttpAuthenticationSecret ` +
+      `FROM ${ENGINE_TABLE} WHERE id = ?`;
+    const results = this.database.prepare(query).get(id);
+
+    if (results.length > 0) {
```
marouanehassanioptimistik, line 82: what if length > 1?

backend/repository/engine.repository.ts
```diff
@@ -0,0 +1,264 @@
...
+      );
+  }
+
+  /**
+   * Create engine settings in the database.
+   * @param {string} id - The ID of the OIBus settings to create. It is not auto generated since it can be shared with other OIBus
+   * @param {EngineSettingsCommandDTO} command - The command
+   */
+  createEngineSettings(id: string, command: EngineSettingsCommandDTO): void {
+    const query =
```
marouanehassanioptimistik, line 206: we might need to check if a line is already there for the given ID and throw an exception or delete the line or use upsert?

marouanehassanioptimistik (commented):
a few comment but the overall look is pretty good

PR #2133, opened 2022-12-17
story-angular-engine-endpoints

burgerni10:
ON TOP OF #2132


marouanehassanioptimistik (commented):
a few comments: 
- is the engine route using the id of the engine as path variable or query param ?
- Do you see the modification of ipfilters, scan mode, proxies as atomic action in the frontend or as part of the whole engine settings
- for scan mode, proxy and ipfilter edition/deletion we need to remember to enforce backend validation rules
- What are external sources, i should have asked in the previous pr

PR #2136, opened 2022-12-23
story-angular-bootstrap

burgerni10:


frontend/src/app/model/types.ts
```diff
@@ -0,0 +1,215 @@
+import { DateTime } from 'luxon';
```
marouanehassanioptimistik, line 1: this seems to contain many useless code :)

frontend/src/app/model/types.spec.ts
```diff
@@ -0,0 +1,15 @@
+import { AGGREGATION_FUNCTION_SETTINGS, ALL_AGGREGATION_FUNCTIONS } from './types';
+
+describe('types', () => {
```
marouanehassanioptimistik, line 3: is that needed ;)?

frontend/tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "noPropertyAccessFromIndexSignature": true,
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
+    "strictNullChecks": false,
```
marouanehassanioptimistik, line 13: is this intended?

marouanehassanioptimistik (changes requested):
i think we should group together all the directives required for form validation FormControlValidation, ValErrorDelayDirective 
@cexbrayat what is the best way to do that ?

cexbrayat:
Maybe we can have an array declared somewhere like this:
```ts
export const formDirectives = [ReactiveFormsModule, FormControlValidation, ...]
```

and then use it in standalone components:
```
standalone: true,
imports: [formDirectives, ...]
```

frontend/src/app/model/types.spec.ts
burgerni10: Indeed I can remove that for now. I added it in preparation of opcua scan groups implementation later.

PR #2146, opened 2023-01-03
feat(proxy): add a proxy list component and retrieve created proxy en…

burgerni10:
image


burgerni10:
> as discussed we should have better semantic in rest api 201 created, 204 no content ...

Yes, It will be done in another PR

PR #2148, opened 2023-01-04
story-angular-generic-south-backend

burgerni10:
ON TOP OF #2147


backend/model/south-connector.model.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * DTO for South connectors
+ */
+export interface SouthConnectorDTO {
+  id: string;
+  name: string;
+  type: string;
+  description: string;
+  enabled: boolean;
+  settings: object;
```
marouanehassanioptimistik, line 10: this doesn't look too good to me, we should have types for settings with inheritance

backend/model/south-connector.model.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Command DTO for South connector
+ */
+export interface SouthConnectorCommandDTO {
+  name: string;
+  type: string;
+  description: string;
+  enabled: boolean;
+  settings: object;
```
marouanehassanioptimistik, line 21: same here

marouanehassanioptimistik (changes requested):

backend/model/south-connector.model.ts
burgerni10: I don't think it will be compatible with a plugin approach ?

burgerni10: I suggest to keep it like that for now, and manage that in a future PR when I'll change the schema on the backend

marouanehassanioptimistik: we should make it compatible i don't see the reason it shouldn't be the plugins should import our types and extend accordingly the base class

marouanehassanioptimistik: if we don't have a model for settings then we have no type safety for south connectors which doesn't seem to good for me. ok to take that into account in a subsequent pr.

PR #2149, opened 2023-01-06
Story angular create south

burgerni10:
ON TOP OF #2155

image

image

image


frontend/src/app/shared/form/form.component.html
```diff
@@ -0,0 +1,16 @@
+<form [formGroup]="form">
+  <div class="row" *ngFor="let row of settingsSchema">
+    <ng-container *ngFor="let settings of row">
+      <div *ngIf="shouldDisplayInput(settings)" [class]="settings.class || 'col'">
```
cexbrayat, line 4: maybe I'm missing something but you don't want to display the disabled inputs right?
so maybe instead of calling `shouldDisplayInput` you could use *ngIf="this.form.get(setting.key).disabled" and remove `shouldDisplayInput`

frontend/src/app/shared/form/oib-checkbox/oib-checkbox.component.ts
```diff
@@ -0,0 +1,24 @@
...
+  templateUrl: './oib-checkbox.component.html',
+  styleUrls: ['./oib-checkbox.component.scss'],
+  viewProviders: [
+    {
+      provide: ControlContainer,
+      useExisting: FormGroupName
+    }
+  ]
+})
+export class OibCheckboxComponent {
```
cexbrayat, line 20: shouldn't that be a ControlValueAccessor?
That would avoid passing the settings as the input

cexbrayat: Same for all other components

frontend/src/app/shared/types.ts
```diff
@@ -0,0 +1,138 @@
...
+  'Australia/Hobart',
+  'Australia/Sydney',
+  'Pacific/Guam',
+  'Australia/Lord_Howe',
+  'Asia/Magadan',
+  'Pacific/Norfolk',
+  'Pacific/Auckland',
+  'Pacific/Fiji',
+  'Pacific/Tongatapu'
+];
```
cexbrayat, line 97: missing `as const`
But instead of declaring them, you can use `Intl.supportedValuesOf('timeZone')`

cexbrayat:
Dans l’ensemble ca a l’air OK, je pense juste que:
- les composants devraient directement etre des control value accessor
- il y a plusieurs itérations sur le schema, je pense que tu devrais arriver à refactorer pour tout faire dans la meme

frontend/src/app/shared/form/form.component.ts
```diff
@@ -0,0 +1,102 @@
...
+            if (settings.conditionalDisplay) {
+              Object.entries(settings.conditionalDisplay).forEach(([key, values]) => {
+                if (key === input) {
+                  if (!values.includes(inputValue)) {
+                    this.form.controls[settings.key].disable();
+                  } else {
+                    this.form.controls[settings.key].enable();
+                  }
+                }
+              });
```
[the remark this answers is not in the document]
burgerni10: I moved the `inputsToSubscribeTo` logic into ngOnInit but I keep this variable to subscribe only once to the form controls (specially if several inputs depends of on another input value)

marouanehassanioptimistik (approved):
no more comment than @cexbrayat please fix them and you can merge

PR #2162, opened 2023-01-19
feat(logs): logs backend repository

burgerni10:


backend/web-server/controllers/log-api.controller.ts
```diff
@@ -0,0 +1,68 @@
...
+const addLogs = async (ctx: KoaContext<LogStreamCommandDTO, void>) => {
+  ctx.request.body.streams.forEach((myStream) => {
+    myStream?.values.forEach((value) => {
+      const formattedLog = {
+        oibus: myStream.stream.oibus,
+        time: new Date(value[0] / 1000000),
+        scope: `${myStream.stream.oibus}:${myStream.stream.scope}`,
+        msg: value[1],
+      };
+      switch (myStream.stream.level) {
```
marouanehassanioptimistik, line 35: shouldn't we have an exhausting switch here to avoid a default case?

burgerni10: I'll fix that with PR #2200 which improves the typescript and eslint rules

PR #2193, opened 2023-01-20
Feature : North Rest API

zhouwang13:
Added a generic North Rest API connector.  The connector allows users to send data to various Rest API with different request body. The connector uses object-mapper to transform the OIBus "values" array into array (or object) using a user-defined mapping.


burgerni10:
We are currently reworking the North connector logic to introduce _transformers_ to adapt the payload to the target application.

I let your PR open for now, so I can come back to you when the transformer logic is validated and implemented. This way you should be able to propose again your connector with this logic that was currently missing, with all the v3 logic.

PR #2197, opened 2023-01-27
Story angular history query edit

burgerni10:
ON TOP OF #2196

### DIFF
--- a/frontend/src/app/history-query/edit-history-query/edit-history-query.component.html
+++ b/frontend/src/app/history-query/edit-history-query/edit-history-query.component.html
@@ -1,0 +1,196 @@
+<div class="oi-padded-container">
+  <div class="d-flex justify-content-between">
+    <div class="oi-header-container">
+      <h1 *ngIf="historyQuery" translate="history-query.edit-title" [translateParams]="{ name: historyQuery.name }"></h1>
+      <h1 *ngIf="!historyQuery" translate="history-query.create-title"></h1>
+    </div>
+  </div>
+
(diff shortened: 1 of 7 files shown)


frontend/src/app/history-query/edit-history-query/edit-history-query.component.html
```diff
@@ -0,0 +1,185 @@
...
+            <oib-scan-mode
+              key="cache-scan-mode"
+              formControlName="scanMode"
+              [scanModes]="scanModes"
+              [label]="'north.cache-settings.send-scan-mode' | translate "
+            ></oib-scan-mode>
+          </div>
+          <!-- Retry interval -->
+          <div class="col-4">
+            <div class="form-group">
```
marouanehassanioptimistik, line 66: i think we are missing the val-errors and it should be tested as well at least for the number of errors expected

frontend/src/app/history-query/edit-history-query/edit-history-query.component.spec.ts
```diff
@@ -0,0 +1,221 @@
...
+              subscriptionOnly: false
+            },
+            settings: []
+          }
+        })
+      );
+      tester = new EditHistoryQueryComponentTester();
+      tester.detectChanges();
+    });
+    it('should display general settings', () => {
```
marouanehassanioptimistik, line 211: nit formatting extra carriage return

marouanehassanioptimistik (changes requested):

PR #2262, opened 2023-04-05
feat(subscription): South subscription in North

kukukk:


burgerni10 (changes requested):
Minor change, otherwise ok

PR #2280, opened 2023-04-13
fix(north): add acceptUnauthorized field in OIAnalytics and OIConnect north connectors

burgerni10:


marouanehassanioptimistik (approved):
Ok we should probably make it clear through a disclaimer in the UI that allowing unauthorized create a vulnerability and should be used with caution as it leaves exposed to a man in the middle attack

PR #2426, opened 2023-06-08
datetime-format

burgerni10:
image

### DIFF
--- a/backend/src/service/utils.spec.ts
+++ b/backend/src/service/utils.spec.ts
@@ -7,7 +7,6 @@ import minimist from 'minimist';
 
 import { DateTime } from 'luxon';
 import * as utils from './utils';
-import { convertDateTimeToInstant, convertDelimiter } from './utils';
 import csv from 'papaparse';
 import pino from 'pino';
 import PinoLogger from '../tests/__mocks__/logger.mock';
--- a/backend/src/service/utils.spec.ts
+++ b/backend/src/service/utils.spec.ts
@@ -7,10 +7,11 @@ import minimist from 'minimist';
 
 import { DateTime } from 'luxon';
 import * as utils from './utils';
+import { convertDateTimeToInstant, convertDelimiter } from './utils';
 import csv from 'papaparse';
 import pino from 'pino';
 import PinoLogger from '../tests/__mocks__/logger.mock';
-import { DateTimeFormat, DateTimeSerialization } from '../../../shared/model/types';
+import { DateTimeFormat } from '../../../shared/model/types';
 
 jest.mock('node:zlib');
 jest.mock('node:fs/promises');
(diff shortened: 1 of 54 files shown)


marouanehassanioptimistik (changes requested):
I think the south model is not precise enough:

Each connector should have an enum for how the date is stored

Mssql => date, datetime or datetime2, smalldatetime, datetimeoffset, varchar
Postgresql => timestamp with tz, timestamp without tz, date, varchar
Mysql => date, datetime, timestamp, varchar
Oracle => DATE, TIMESTAMP, TIMESTAMP WITH TIME ZONE, and TIMESTAMP WITH LOCAL TIME ZONE, varchar
sqllite => text
ip21 => ?

If the datatype chosen is not a data type that supports timezone than a timezone should be required by the connector to timezone the retrieved value

If the datatype chosen is not a native date format then a format should be mandatory to represent the format of the timestamp in the db (different then the format use for serialization)

Once the retrieval is made and instants are build from the retrieved value lastCompleted should be stored in utc and natively binded to the query using the appropriate format according to the column (no need for extra config I think)

For serializing the value into the file another timestamp format (as well as a timezone) is required but it should be clear this one is only used for serialization.

burgerni10:
> I think the south model is not precise enough:
> 
> Each connector should have an enum for how the date is stored
> 
> Mssql => date, datetime or datetime2, smalldatetime, datetimeoffset, varchar Postgresql => timestamp with tz, timestamp without tz, date, varchar Mysql => date, datetime, timestamp, varchar Oracle => DATE, TIMESTAMP, TIMESTAMP WITH TIME ZONE, and TIMESTAMP WITH LOCAL TIME ZONE, varchar sqllite => text ip21 => ?
> 
> If the datatype chosen is not a data type that supports timezone than a timezone should be required by the connector to timezone the retrieved value
> 
> If the datatype chosen is not a native date format then a format should be mandatory to represent the format of the timestamp in the db (different then the format use for serialization)
> 
> Once the retrieval is made and instants are build from the retrieved value lastCompleted should be stored in utc and natively binded to the query using the appropriate format according to the column (no need for extra config I think)
> 
> For serializing the value into the file another timestamp format (as well as a timezone) is required but it should be clear this one is only used for serialization.

- MSSQL: I've managed to include the specific datetime object as a tuning parameter when 'date-object' is selected
- MySQL: MySQL store everything in UTC, but change [the timezone depending on the connection settings](https://dev.mysql.com/doc/refman/8.0/en/datetime.html). OIBus already set the timezone to "Z". That means the timezone is unnecessary for this driver. Moreover the driver convert [Date objects into 'YYYY-mm-dd HH:ii:ss' strings](https://github.com/mysqljs/mysql). That allows the server to compare date, timestamp and Datetime alike.
- PostgreSQL: OIBus already have a converter on the driver with this line : ``pg.types.setTypeParser(1114, str => new Date(`${str}Z`));``. 1114 stands for TIMESTAMP type. That means all datetime without timezone are converted into UTC timestamp. I think it is wrong and should be refactor. Let's discuss that. [The date-object is converted by the library to a UTC string ](https://node-postgres.com/features/queries). This means we don't need to add date-object subtypes for postgresql
- Oracle: [this documentation](https://node-oracledb.readthedocs.io/en/latest/user_guide/sql_execution.html#fetching-dates-and-timestamps) suggest that we only have to manage JS DateTime object, no need to specify sub-types. However, we may have something tricky to remove : ``process.env.ORA_SDTZ = 'UTC';      oracledb.outFormat = oracledb.OUT_FORMAT_OBJECT;``. I suggest to dig into that further in another issue
- ODBC: no additional types needed
- SQLite: no additional types needed
- IP21: no additional types needed

PR #2435, opened 2023-06-13
test-connection-button

burgerni10:


backend/src/south/south-interface.ts
```diff
@@ -1,5 +1,5 @@
 import pino from 'pino';
-import { OibusItemDTO, SouthConnectorDTO } from '../../../shared/model/south-connector.model';
+import { OibusItemDTO, SouthConnectorCommandDTO, SouthConnectorDTO } from '../../../shared/model/south-connector.model';
```
nagyszabi, line 2: `SouthConnectorDTO` is unused

PR #2445, opened 2023-06-15
Feature: Add SQLite connection test

nagyszabi:


backend/src/south/south-sqlite/south-sqlite.ts
```diff
@@ -56,10 +57,24 @@ export default class SouthSQLite extends SouthConnector implements QueriesHistor
...
+      logger.trace(`Access error on '${dbFolder}': ${error.message}`);
+      throw new Error(`Folder '${dbFolder}' does not exist`);
+    }
+
+    try {
+      await fs.access(dbFolder, fs.constants.R_OK | fs.constants.W_OK);
+    } catch (error: any) {
+      logger.trace(`Access error on '${dbFolder}': ${error.message}`);
+      throw new Error(`No read/write access on folder`);
+    }
```
burgerni10, line 77: I think we could improve this test a little bit more by querying the sqlite_master table : `SELECT * from sqlite_master` and log.info the result.

It should output something like on this screenshot and give useful info for debugging / testing / writing queries
image

By the way, we should query a system table on other sql connectors too, in order to test the connection properly

burgerni10 (changes requested):

backend/src/south/south-sqlite/south-sqlite.ts
nagyszabi: Alright, and should we throw an error if the specified db file is not present? Because by default, the sqlite package creates the file if it's not present.

burgerni10: Yes, let's first check if the file exists. If not, throw an error, if yes, request the database

burgerni10 (approved):

PR #2500, opened 2023-07-03
Feature: Add ODBC connection test

nagyszabi:


burgerni10 (approved):
I have a more general comment on the "static" part of the connection tests. Because of the static feature, we have to make everything we use static.

I think we should remove the static method for a class method, and instantiate a class instance in the controller. Can you try something like that in another PR ?

Issue #2279, opened 2023-04-13
[North] - Retry files (from error) do not update the queue


burgerni10:
When retrying files, update the queue of the north connector

PR #2522, opened 2023-07-17
Fix: Retry files (from error) do not update the queue

nagyszabi:
Fixes: #2279


frontend/src/app/north/explore-cache/explore-cache.component.html
nagyszabi, line 1: I added this check because when the component would instantiate, it threw an error that `northConnector` is `null`

burgerni10 (approved):

PR #2543, opened 2023-07-20
fix(styles): change of south/north/history-query icon

burgerni10:
add new files:
history-query.svg + north.svg + south.svg.

New class: oib-title-image/title-image
(the purpose of these css classes is to resize the new icons)

correct bug about oib-box


frontend/src/app/history-query/history-query-list.component.scss
burgerni10, line 1: This code is dupplicated. Please note that the file styles.scss allows you to share your style across components.

Also, don't redefine bootstrap style in specific component : use it as it is or create new css classes (starting with the oi- prefix so we know that it's a custom style). If this class is used only in this component, it's ok to put it in this scss file, otherwise it goes in the main scss. 

If you want to change a global style (like h1), do that in the main styles.scss file

frontend/src/app/shared/box/box.component.html
```diff
@@ -4,7 +4,10 @@
       <ng-container *ngIf="titleTemplateRef; else stringTitle">
         <ng-container *ngTemplateOutlet="titleTemplateRef"></ng-container>
       </ng-container>
-      <ng-template #stringTitle>{{ boxTitle | translate }}</ng-template>
+      <ng-template #stringTitle>
+        <img class="oib-title-image" *ngIf="imagePath" [src]="imagePath" alt="Image" /> <!-- Affiche l'image correspondante si imagePath n'est pas vide -->
```
burgerni10, line 8: Don't put french comments: an open source project (and an IT project in general) may include non french speaking people, so this comment will be more difficult to handle for them

frontend/src/app/shared/box/box.component.ts
```diff
@@ -37,6 +37,7 @@ export class BoxTitleDirective {
 export class BoxComponent implements AfterContentInit {
   @Input() boxTitle = '';
   titleTemplateRef: TemplateRef<void> | null = null;
+  imagePath = '';
```
burgerni10, line 40: Use a non-mandatory  input instead of checking the title of the box to know what to display. It will be much more modular this way

burgerni10 (commented):
Several changes are necessary before merging. The icons themselves are ok, but the code quality is... less ok ;) 

Please use prettier or run `npm run lint -- --fix` to format your files correctly, so we use the same standards

Issue #2272, opened 2023-04-11
[North] - Allow subscription to external sources


burgerni10:
North connectors can subscribe to either south connectors or external sources (when data come from an endpoint, we filter by the 'name' query param). For now, external sources are not taken into account.

This issue should solve that:
- [x] On the engine by filtering north when adding values / files into the cache
- [x] On the repository by adding one table associating north connectors with south connectors and another table associating north connectors with external sources
- [x] On the API controller by adding/deleting and getting the subscriptions for each north (for other south and for external sources)

PR #2548, opened 2023-07-24
Issue #2272: Subscribe to external sources

kukukk:


backend/src/web-server/controllers/oibus.controller.ts
```diff
@@ -59,8 +64,13 @@ export default class OibusController extends AbstractController {
     const { name } = ctx.request.query;
     const file = ctx.request.file;
     if (name && file) {
+      const externalSource = ctx.app.repositoryService.externalSourceRepository.findExternalSourceByReference(name as string);
+      if (!externalSource) {
+        return ctx.notFound();
+      }
+
```
burgerni10, line 71: The default behaviour is to store the data, even if the external source is missing. This way, we can still have data sent to connectors without subscription, but keeping trace of the external source in the log. 
I think we should keep it like this behaviour if possible ?

backend/src/web-server/routes/index.ts
```diff
@@ -94,6 +94,13 @@ router.post('/api/north/:northId/subscriptions/:southId', (ctx: KoaContext<any,
 router.delete('/api/north/:northId/subscriptions/:southId', (ctx: KoaContext<any, any>) =>
   subscriptionController.deleteNorthSubscription(ctx)
 );
+router.get('/api/north/:northId/external-subscriptions', (ctx: KoaContext<any, any>) => subscriptionController.getExternalNorthSubscriptions(ctx));
+router.post('/api/north/:northId/external-subscriptions/:southName', (ctx: KoaContext<any, any>) =>
```
burgerni10, line 98: To have some consistency, southName should be renamed reference: it is not always sent by a south, but it can be sent directly from another application through the OIBus API, so the reference field refers to this source (another OIBus or another app)

burgerni10 (approved):

PR #2567, opened 2023-07-26
fix-values-compact

burgerni10:


marouanehassanioptimistik (approved):
LGTM it would be useful before deploying to test a stop restart just to make sure everything is fine

burgerni10:
> LGTM it would be useful before deploying to test a stop restart just to make sure everything is fine

After a test with a 5GB cache, the task manager shows a 50% peak of cpu usage for 3/4s on a 2,5GHz processor. Without this change, it takes several minutes before the North is fully started with all file scanned. 

 I think it will improve startup performances quite a lot

PR #2629, opened 2023-08-22
feat(engine-north): add certificates to engine and azure active directory cert mgt in north OIA

burgerni10:


backend/src/db/entity-migrations/v3.1.1-add-certificates-table.ts
```diff
@@ -0,0 +1,24 @@
...
+  await knex.schema.raw(`create table ${CERTIFICATES_TABLE}
+                                  (id          char(36) primary key,
+                                   created_at  datetime default CURRENT_TIMESTAMP not null,
+                                   updated_at  datetime default CURRENT_TIMESTAMP not null,
+                                   name        varchar not null,
+                                   description varchar,
+                                   public_key varchar not null,
+                                   private_key varchar not null,
+                                   private_key_passphrase varchar,
+                                   expiry varchar,
```
burgerni10, line 18: Sometimes, it is useful to add the CA associated with the certificate, (see south MQTT for example)
Should we add an optional field for that ?

shared/model/certificate.model.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export interface CertificateOptions {
+  commonName: string;
+  countryName: string;
+  stateOrProvinceName: string;
+  localityName: string;
+  organizationName: string;
+  keySize: number;
+  daysBeforeExpiry: number;
```
burgerni10, line 34: Shouldn't we have a private_key_passphrase field (stored as OibSecret) ?

frontend/src/app/shared/clipboard-copy-directive.ts
```diff
@@ -0,0 +1,17 @@
+import { Directive, HostListener, Input } from '@angular/core';
+import { Clipboard } from '@angular/cdk/clipboard';
+
+@Directive({
+  selector: '[oibClipboardCcopy]',
```
burgerni10, line 5: oibClipboardCcopy -> oibClipboardCopy

shared/model/certificate.model.ts
marouanehassanioptimistik: i hesitated but i don't see the use. private key is stored encrypted in the database so the only use case would be to download the private key in a encrypted mode and use a passphrase for that. But at certificate generation it does seem redundant. Do you agree?

backend/src/db/entity-migrations/v3.1.1-add-certificates-table.ts
marouanehassanioptimistik: how is a CA expressed in the form of another certificate stored in pem format?
do we need a root CA as well or a concatenation of both?

burgerni10: When authenticating for MQTT with cert, the root CA is needed alongside the auth certificate. 
Other use cases (such as MongoDB I think) use a different approach where we need to concatenate both in one single cert.

That's why I suggest an optional field with root CA for these cases. That being said, it can be done later as an enhancement where the certificate service will be used for other protocols (such as MQTT, OPCUA...)

marouanehassanioptimistik: but then that would not be for self signed certificate would it. It would be for manually imported certificate where we don't have a service and ui yet.

shared/model/certificate.model.ts
burgerni10: Indeed it will be useful only when downloading the private key, and to protect it with a passphrase. But we can do it by design by protecting the private key with a passphrase that will be used inside OIBus at run time AND to protect the private key by default when downloading it.

Anyway, it can also be an improvement, but the private_key_passphrase field in the database can then be removed for now, since it's never used

marouanehassanioptimistik: ha yeah it was a leftover in the knex migration i intended to remove it, it is done

nicolasnoel (approved):
It works !

Issue #2614, opened 2023-08-14
[Cache] - Update cache when changing scan mode or max instant per item


burgerni10:
On South connectors, the cache is not updated when changing scan mode for an item, or when changing from maxInstantPerItem = true to false (on OPCUA for example).

Here is an example:
- maxInstant is 2023-01-01T00:00:00Z for scan mode every10seconds
- The user changes to another scan mode (every1second) so the new maxInstant associated to this scan mode is set to the last max instant (eg 2023-01-01T01:00:00Z) but the max instant associated to the scan mode every10seconds is still 2023-01-01T00:00:00Z
- One month later, the user goes back to every10seconds. On next trigger, it will request the data from 2023-01-01T00:00:00Z even though we retrieved the data until now

To fix that, I suggest that when changing the scan mode for an item, the last max instant is removed (if the scan mode is not used for another items with maxInstantPerItem = false)

Moreover, when switching maxInstantPerItem from true to false and vice versa, the same problem may occur, so it must also remove the last max instant

If a last max instant is found, the new max instant may be set to the last max instant, to have continuity.

PR #2760, opened 2023-08-29
fix(cache): Update cache on scan mode or max instant per item change

nagyszabi:
Fixes: #2614


backend/src/service/reload.service.ts
```diff
@@ -110,6 +112,23 @@ export default class ReloadService {
...
+
+      for (const item of southItems) {
+        this.repositoryService.southCacheRepository.createOrUpdateCacheScanMode({
+          southId,
+          itemId: command.history.maxInstantPerItem ? item.id : 'all', // When max instant per item is disabled, we use 'all' as item_id
+          scanModeId: item.scanModeId,
+          maxInstant: maxInstants.get(item.scanModeId) || latestInstant
+        });
+      }
+    }
```
burgerni10, line 131: This should be called before the if(command.enabled) line because if south connector restart, it will take the previous maxInstant

backend/src/service/reload.service.ts
```diff
@@ -152,6 +171,22 @@ export default class ReloadService {
...
+      this.repositoryService.southCacheRepository.updateCacheScanModeId(southId, newItem.id, oldScanModeId, newScanModeId);
+
+      if (!settings.history.maxInstantPerItem) {
+        const southItems = this.repositoryService.southItemRepository.getSouthItems(southId);
+        const isOldScanModeUnused = !southItems.some(item => item.scanModeId === oldScanModeId);
+        if (isOldScanModeUnused) {
+          this.repositoryService.southCacheRepository.deleteCacheScanMode(southId, oldScanModeId, 'all');
+        }
+      }
+    }
```
burgerni10, line 189: This should be called before the `this.oibusEngine.updateItemInSouth(southId, southItem, newItem);` line because it may restart right away the item with the old max instant

burgerni10 (changes requested):

burgerni10 (approved):

PR #2811, opened 2023-09-08
feat(north): Add north item mode and interface

burgerni10:


nagyszabi:
I've rebased the branch with main @burgerni10 🙂

backend/src/db/entity-migrations/v3.5.0-add-transformers.ts
```diff
@@ -0,0 +1,48 @@
...
+  });
+}
+
+async function createHistoryTransformersTable(knex: Knex): Promise<void> {
+  await knex.schema.createTable(HISTORY_TRANSFORMERS_TABLE, table => {
+    table.uuid('history_id');
+    table.uuid('north_transformer_id');
+    table.uuid('south_transformer_id');
+  });
+}
```
burgerni10, line 46: All these uuid should be foreign keys

backend/src/web-server/controllers/north-connector.controller.ts
```diff
@@ -539,4 +546,274 @@ export default class NorthConnectorController {
...
+    ctx.ok();
+  }
+
+  async checkImportNorthItems(ctx: KoaContext<{ itemIdsToDelete: string }, any>): Promise<void> {
+    const manifest = this.getManifestWithItemsMode(ctx);
+
+    const file = ctx.request.file;
+    if (file.mimetype !== 'text/csv') {
+      return ctx.badRequest();
+    }
```
burgerni10, line 622: I actually had to remove this because on some Windows system, it was not working properly. Let's discuss that

backend/src/db/entity-migrations/v3.5.0-add-transformers.ts
```diff
@@ -0,0 +1,79 @@
...
+    table.string('file_regex');
+  });
+}
+
+async function createNorthTransformersTable(knex: Knex): Promise<void> {
+  await knex.schema.createTable(NORTH_TRANSFORMERS_TABLE, table => {
+    table.uuid('north_id').notNullable().references('id').inTable(NORTH_CONNECTORS_TABLE);
+    table.uuid('transformer_id').notNullable().references('id').inTable(TRANSFORMERS_TABLE);
+
+    table.unique(['north_id', 'transformer_id']);
```
burgerni10, line 50: Not sure this should be unique: we could have transformers associated to a regex in a north connector to parse specific files and another transformer with another regex

backend/src/repository/history-transformer.repository.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    if (!filter) {
+      const query = `SELECT t.id, t.name, t.description, t.input_type AS inputType, t.output_type AS outputType, t.code, t.file_regex AS fileRegex FROM ${HISTORY_TRANSFORMERS_TABLE} ht
+        JOIN transformers t ON ht.transformer_id = t.id
+        WHERE ht.history_id = ? AND ht.connector_type = ?;`;
+      return this.database.prepare(query).all(historyId, connectorType) as Array<TransformerDTO>;
+    }
+
+    let query = `SELECT t.id, t.name, t.description, t.input_type AS inputType, t.output_type AS outputType, t.code, t.file_regex AS fileRegex FROM ${HISTORY_TRANSFORMERS_TABLE} ht
+      JOIN transformers t ON ht.transformer_id = t.id`;
```
burgerni10, line 38: Shouldn't we use the TRANSFORMERS_TABLE constant instead?

nagyszabi: Yes, will do

backend/src/db/entity-migrations/v3.5.0-add-transformers.ts
nagyszabi: In that situation we would have two different transformers:
- `transformer_1` with regex
- `transformer_2` without 

and the table would let the values:
`north_1`, `transformer_1`
`north_1`, `transformer_2`

is my understanding correct?

PR #2837, opened 2023-09-28
Fix(north): Fix import / export north items

nagyszabi:
Applies the fixes done to the importing and exporting of south items (#2826 and #2838) for north items


burgerni10 (changes requested):
I've made some changes (again) to fix a few issues regarding the export / import of south items. 
Can you also fix the #2841 (from main) and take these changes into account in this PR ?

burgerni10 (approved):

Issue #2890, opened 2023-10-02
[North] - Retrieve regular cache files


burgerni10:
Add methods to retrieve regular cache files in a new section on the cache page. Add methods to delete a regular cache file, or place it in the archive folder.

PR #2960, opened 2023-10-17
feat(north-cache): Show regular cache files

nagyszabi:
Closes: #2890


burgerni10 (changes requested):
When removing or archiving files, the queue `filesQueue` in file-cache.service.ts must be updated by filtering out the file to remove/archive.

burgerni10 (approved):

Issue #2891, opened 2023-10-02
[History Query] - Add connection test (south / north parts)


burgerni10:
Add connection tests in History query. 

Two endpoints should be implemented : one for testing the south connection, one for testing the north connection.

The same logic than the south/north connection tests can then be applied

PR #2981, opened 2023-10-31
Feature: Add connector test to history queries

nagyszabi:
Closes: #2891

### DIFF
--- a/backend/src/web-server/controllers/history-query.controller.ts
+++ b/backend/src/web-server/controllers/history-query.controller.ts
@@ -3,6 +3,8 @@ import { HistoryQueryCommandDTO, HistoryQueryCreateCommandDTO, HistoryQueryDTO }
 import JoiValidator from './validators/joi.validator';
 
 import {
+  SouthConnectorCommandDTO,
+  SouthConnectorDTO,
   SouthConnectorItemCommandDTO,
   SouthConnectorItemDTO,
   SouthConnectorItemSearchParam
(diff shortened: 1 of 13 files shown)


burgerni10 (changes requested):
Can you add tests to keep the cover at 100% on the backend ? Otherwise, it looks good !

backend/src/web-server/controllers/history-query.controller.ts
```diff
@@ -439,4 +442,98 @@ export default class HistoryQueryController extends AbstractController {
...
+      let southSettings: SouthConnectorDTO['settings'] | null = null;
+      // When we have a history query id, we can retrieve the old connector settings from repository
+      if (ctx.params.id !== 'create') {
+        southSettings = ctx.app.repositoryService.historyQueryRepository.getHistoryQuery(ctx.params.id)?.southSettings;
+        if (!southSettings) {
+          return ctx.notFound();
+        }
+      }
+      // When creating a new history query from exising connector, we can retrieve old connector settings using connector id
+      else if (ctx.request.query?.fromConnectorId && typeof ctx.request.query?.fromConnectorId === 'string') {
```
nagyszabi, line 464: I've updated the tests but I couldn't get them to 100%.

At this `else if` (both here and for north test) seems to be the issue. <br>
I have tests that check cases where `fromConnectorId` isn't set, and where it's not a string. I think the issue is with the string checking.

Could you take a look as well? Or do you have any suggestions regarding this?

burgerni10 (approved):

PR #3052, opened 2023-11-13
chore: use application builder

cexbrayat:


cexbrayat:
Don't merge for now until we figure out the i18n issue we spotted on OiA

PR #3406, opened 2024-05-10
north-handle-content

burgerni10:


backend/src/north/north-oibus/north-oibus.ts
```diff
@@ -71,10 +70,22 @@ export default class NorthOibus extends NorthConnector<NorthOIBusSettings> imple
...
 
+  async handleContent(data: OIBusContent): Promise<void> {
+    switch (data.type) {
+      case 'raw':
+        await this.handleFile(data.filePath);
+        return;
+
+      case 'time-values':
+        await this.handleValues(data.content);
+        break;
```
nagyszabi, line 81: These can be written simply as `return this.handleFile(data.filePath);` since there is no need to await them. It's alright to just return the promise, because it will be awaited when the `handleContent` function is called: https://github.com/OptimistikSAS/OIBus/blob/94b785fd29575fdf3ffc12235f48c4ac344313ea/backend/src/north/north-connector.ts#L275

PR #3410, opened 2024-05-13
add-content

burgerni10:


backend/src/south/south-mqtt/south-mqtt.ts
```diff
@@ -148,29 +147,38 @@ export default class SouthMQTT extends SouthConnector<SouthMQTTSettings, SouthMQ
...
+                timestamp: messageTimestamp,
+                data: {
+                  value: message.toString()
+                }
               }
-            }
-          ]);
+            ]
+          });
           break;
```
nagyszabi, line 176: Since these two cases are the same, they can be shortened as:
```ts
        case 'number':
        case 'string':
          await this.addContent({
            type: 'time-values',
            content: [
              {
                pointId: associatedItem.name,
                timestamp: messageTimestamp,
                data: {
                  value: message.toString()
                }
              }
            ]
          });
          break;
```

backend/src/web-server/controllers/oibus.controller.ts
```diff
@@ -57,41 +57,59 @@ export default class OibusController extends AbstractController {
   }
 
   async addValues(ctx: KoaContext<void, void>): Promise<void> {
```
nagyszabi, line 59: Here `addValues` and `addFile` are two separate methods called here:
https://github.com/OptimistikSAS/OIBus/blob/f1d62c74f1aeacb69f729cd7df823ba88530e2e0/backend/src/web-server/routes/index.ts#L88-L89

Maybe we make it into one `/api/add-external-content` endpoint, with a query param of `type`?
This way, we have the flexibility of adding new types without much code duplication.

PR #3520, opened 2024-07-12
engine-reload-settings

burgerni10:


backend/src/engine/oibus-engine.ts
```diff
@@ -163,51 +175,41 @@ export default class OIBusEngine extends BaseEngine {
...
+    }
 
-  async updateItemInSouth(southId: string, oldItem: SouthConnectorItemDTO, newItem: SouthConnectorItemDTO): Promise<void> {
-    await this.southConnectors.get(southId)?.updateItem(oldItem, newItem);
+    // Do not await here, so it can start all connectors without blocking the thread
+    north.start().catch(error => {
+      this.logger.error(
+        `Error while starting North connector "${north.getSettings().name}" of type "${north.getSettings().type}" (${north.getSettings().id}): ${error}`
+      );
+    });
```
nagyszabi, line 194: This is called by the reload service here:
https://github.com/OptimistikSAS/OIBus/blob/650c83a3f23ccea1ca7d363525d373d6dc6a2a42/backend/src/service/reload.service.ts#L278-L281

Which sets the db status as `enabled` before the `this.oibusEngine.startNorth` call.
But if the `north.start()` call fails, the database will still have the `enabled` status set to true.

So the engine should handle this somehow, to have the `enabled` status set to false if the `north.start()` call fails.
This seems to be a bit hard to implement at the moment, because it would require to pass down some sort of `onErrorFn` function to the engine, which should come from the reload service.

backend/src/engine/history-query-engine.ts
```diff
@@ -69,28 +66,18 @@ export default class HistoryQueryEngine extends BaseEngine {
...
-    await this.historyQueries.get(historyId)?.deleteItem(item);
-  }
-
-  async deleteAllItemsFromHistoryQuery(historyId: string): Promise<void> {
-    await this.historyQueries.get(historyId)?.deleteItems();
-  }
-
-  async updateItemInHistoryQuery(historyId: string, item: SouthConnectorItemDTO): Promise<void> {
-    await this.historyQueries.get(historyId)?.updateItem(item);
+  async startHistoryQuery(historySettings: HistoryQueryDTO): Promise<void> {
```
nagyszabi, line 72: Just to have the same code logic as for the oibus engine, in the creation and starting of things, it is enough to pass the `historyId` to the start function. This way, `create` needs settings, and `start` needs only an id, just like how the connectors do it:
https://github.com/OptimistikSAS/OIBus/blob/650c83a3f23ccea1ca7d363525d373d6dc6a2a42/backend/src/engine/oibus-engine.ts#L134
https://github.com/OptimistikSAS/OIBus/blob/650c83a3f23ccea1ca7d363525d373d6dc6a2a42/backend/src/engine/oibus-engine.ts#L149

backend/src/engine/history-query-engine.ts
```diff
@@ -69,28 +66,18 @@ export default class HistoryQueryEngine extends BaseEngine {
       this.historyQueries.delete(settings.id);
       this.historyQueries.set(settings.id, historyQuery);
     }
-    this.historyQueries
-      .get(settings.id)!
-      .start()
-      .catch(error => {
-        this.logger.error(error);
-      });
+    await this.startHistoryQuery(settings);
```
nagyszabi, line 69: `startHistoryQuery` will actually be called twice, if the user creates a new history query:
1. First of all, the user calls the endpoint to create the query, which calls the reload service:
https://github.com/OptimistikSAS/OIBus/blob/650c83a3f23ccea1ca7d363525d373d6dc6a2a42/backend/src/web-server/controllers/history-query.controller.ts#L128
2. The reload service then creates and starts the HQ as well:
https://github.com/OptimistikSAS/OIBus/blob/650c83a3f23ccea1ca7d363525d373d6dc6a2a42/backend/src/service/reload.service.ts#L318-L319
3. But because `createHistoryQuery` calls `startHistoryQuery`:
_see code refferenced in this comment_
4. The `startHistoryQuery` gets called a second time, in the reload service:
_see point 2._

This can be fixed in two ways, either by removing the `startHistoryQuery` from `createHistoryQuery`, or from the reload service. I think removing it from the reload service might be the good idea, but I'm not 100% sure.

nagyszabi (changes requested):
All in all looks very much better, and it really clicked why moving the item management into the connectors themselves improves the code. 😄 
Looks cleaner and the whole mental model of how things work is easier now.
I found some bugs but overall looks very good 💯

backend/src/engine/history-query-engine.ts
nagyszabi: I just noticed that it makes the reload service code easier as well:
https://github.com/OptimistikSAS/OIBus/blob/650c83a3f23ccea1ca7d363525d373d6dc6a2a42/backend/src/service/reload.service.ts#L368-L373

Here for example, there is no need to call `const settings = this.repositoryService.historyQueryRepository.getHistoryQuery(historyId)!;` anymore, we can simply pass `historyId` to the start method.

backend/src/south/south-connector.ts
```diff
@@ -439,18 +502,18 @@ export default class SouthConnector<T extends SouthSettings = any, I extends Sou
...
-      cronJob.stop();
-    }
-    this.cronByScanModeIds.clear();
     this.logger.debug(`Stopping South "${this.connector.name}" (${this.connector.id})...`);
 
+    if (dataStream) {
+      // Reload the settings only on data stream case, otherwise let the history query manage the settings
+      this.connector = this.repositoryService.southConnectorRepository.getSouthConnector(this.connector.id)!;
+    }
+
```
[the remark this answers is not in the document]
burgerni10: There is an edgy case where this is needed: when a run is in progress, the settings can be changed by the user. If the run fails because of connection failure, it sometimes reset the connection, like in South-OPCUA.

https://github.com/OptimistikSAS/OIBus/blob/650c83a3f23ccea1ca7d363525d373d6dc6a2a42/backend/src/south/south-opcua/south-opcua.ts#L336-L338

If the settings are not reloaded, it will try to reconnect again. Hence this update here.

Issue #3414, opened 2024-05-17
OIBus -> OIAnalytics Registeration, add proxy configuration without having to unregister


brunomaury:
Scenario : 
1. There is no proxy, we register OIAnalytics in OIBus and add the corresponding North
2. A proxy is added, we want to add the proxy configuration in OIBus
Currently, we have to first unregister OIAnalytics in OIBus.
But, since the registeration through the proxy seems more complicated (or maybe a bit bugged), it's not convenient.

PR #3523, opened 2024-07-15
fix: add proxy configuration without unregister

ceedjy:
fix: #3414

### DIFF
--- a/frontend/src/app/services/engine.service.ts
+++ b/frontend/src/app/services/engine.service.ts
@@ -66,6 +66,13 @@ export class EngineService {
     return this.http.put<void>(`/api/registration`, command);
   }
 
+  /**
+   * Edit the engine settings
+   */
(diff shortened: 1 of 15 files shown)


frontend/src/app/services/engine.service.ts
```diff
@@ -66,6 +66,13 @@ export class EngineService {
     return this.http.put<void>(`/api/registration`, command);
   }
 
+  /**
+   * Edit the engine settings
+   */
+  editRegistrationSettings(command: RegistrationSettingsCommandDTO): Observable<void> {
+    return this.http.put<void>(`/api/registration/edit`, command);
+  }
```
burgerni10, line 74: You forgot to test this method

burgerni10 (changes requested):

burgerni10 (approved):

PR #3543, opened 2024-07-25
feat: export items delimiter

ceedjy:
fix: #233

### DIFF
--- a/frontend/src/app/services/history-query.service.ts
+++ b/frontend/src/app/services/history-query.service.ts
@@ -176,10 +176,10 @@ export class HistoryQueryService {
   /**
    * Export items in CSV file
    */
-  exportItems(historyQueryId: string, historyQueryName: string): Observable<void> {
+  exportItems(historyQueryId: string, fileName: string, delimiter: string): Observable<void> {
     return this.http
-      .get(`/api/history-queries/${historyQueryId}/south-items/export`, { responseType: 'blob', observe: 'response' })
-      .pipe(map(response => this.downloadService.download(response, `${historyQueryName}-south-items.csv`)));
+      .put(`/api/history-queries/${historyQueryId}/south-items/export`, { delimiter }, { responseType: 'blob', observe: 'response' })
+      .pipe(map(response => this.downloadService.download(response, `${fileName}.csv`)));
   }
 
   /**
@@ -203,7 +203,8 @@ export class HistoryQueryService {
   checkImportItems(
     southType: string,
     historyQueryId: string,
-    file: File
+    file: File,
+    delimiter: string
   ): Observable<{
     items: Array<SouthConnectorItemDTO>;
     errors: Array<{
@@ -213,6 +214,7 @@ export class HistoryQueryService {
   }> {
     const formData = new FormData();
     formData.set('file', file);
+    formData.set('delimiter', delimiter);
     return this.http.post<{ items: Array<SouthConnectorItemDTO>; errors: Array<{ item: SouthConnectorItemDTO; message: string }> }>(
       `/api/history-queries/${southType}/south-items/check-south-import/${historyQueryId}`,
       formData
--- a/frontend/src/app/shared/export-item-modal/export-item-modal.component.html
+++ b/frontend/src/app/shared/export-item-modal/export-item-modal.component.html
@@ -1,0 +1,27 @@
+<div class="modal-header">
+  <h4 class="modal-title">Export option</h4>
+</div>
+<div class="modal-body">
+  <div>
+    <div [formGroup]="exportForm" class="form-group">
+      <div>
+        <label class="col form-label" for="delimiter" translate="common.delimiter"></label>
+        <select formControlName="delimiter" class="form-select w-50 mb-3" id="delimiter">
+          @for (delimiter of csvDelimiters; track delimiter) {
+            <option [ngValue]="delimiter">{{ 'enums.csv-character.' + delimiter | translate }}</option>
+          }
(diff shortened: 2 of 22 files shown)


frontend/src/app/services/history-query.service.ts
```diff
@@ -219,6 +219,16 @@ export class HistoryQueryService {
...
 
+  /**
+   * Ckeck correspondence between the delimiter entered and the delimiter in the CSV file
+   */
+  checkCorrespondence(file: File, delimiter: string) {
+    const formData = new FormData();
+    formData.set('file', file);
+    formData.set('delimiter', delimiter);
+    return this.http.post<{ correspond: boolean }>(`/api/history-queries/items/check-correspondence`, formData);
+  }
```
burgerni10, line 230: I'm not sure we need this : for export, we just use a delimiter from a list, and for the import, we just throw an error when checking the file, and we display the error on the frontend.

burgerni10: Same remark for south-connector

frontend/src/app/shared/export-item-modal/export-item-modal.component.html
```diff
@@ -0,0 +1,19 @@
...
+        <option [value]="option">{{ option }}</option>
+      }
+    </select>
+  </div>
+</div>
+<div class="modal-footer">
+  <div class="btn-group">
+    <button type="button" class="btn btn-primary" (click)="save()" id="save-button">Save</button>
+    <button type="button" class="btn btn-cancel" (click)="cancel()" id="cancel-button">Cancel</button>
+  </div>
```
burgerni10, line 18: Please use internationalisation with a translation key

frontend/src/app/shared/export-item-modal/export-item-modal.component.html
```diff
@@ -0,0 +1,19 @@
+<div class="modal-header">
+  <h4 class="modal-title">Export option</h4>
+</div>
+<div class="modal-body">
+  <div>
+    <label class="col form-label" for="delimiter">Delimiter</label>
+    <select name="delimiter" class="col form-select w-50">
+      @for (option of delimiterOptions; track option) {
+        <option [value]="option">{{ option }}</option>
+      }
```
burgerni10, line 10: You should use an enum translation (see other example in the code) to display "Semi-Colon ;" instead of just ";"

frontend/src/app/shared/export-item-modal/export-item-modal.component.ts
```diff
@@ -0,0 +1,28 @@
...
+  standalone: true
+})
+export class ExportItemModalComponent {
+  delimiterOptions: Array<string> = [',', ';'];
+  selectedDelimiter: string = ',';
+
+  constructor(private modal: NgbActiveModal) {}
+
+  save() {
+    const delimiter = document.querySelector('select')?.value;
```
burgerni10, line 18: you should use a form control field (see other forms inputs of type select). Also, it should be required, so it's not useful to check if it exists

frontend/src/app/shared/import-item-modal/import-item-modal.component.html
```diff
@@ -0,0 +1,33 @@
+<div class="modal-header">
+  <h4 class="modal-title">Import Items</h4>
```
burgerni10, line 2: Use internationalisation

frontend/src/app/shared/import-item-modal/import-item-modal.component.html
```diff
@@ -0,0 +1,33 @@
...
+</div>
+<div class="modal-body">
+  <div class="row">
+    <label for="file">Choose a file</label>
```
burgerni10, line 6: Use internationalisation

frontend/src/app/shared/import-item-modal/import-item-modal.component.html
```diff
@@ -0,0 +1,33 @@
...
+      (click)="fileInput.click()"
+      (drop)="onImportDrop($event)"
+      (dragover)="onImportDragOver($event)"
+      id="import-button"
+    >
+      <span>{{ this.selectedFile.name }}</span>
+    </button>
+  </div>
+  <div>
+    <label class="col form-label" for="delimiter">What is the delimiter ?</label>
```
burgerni10, line 19: Use internationalisation

burgerni10: Also, instead of a question, juste write "Delimter"

frontend/src/app/shared/import-item-modal/import-item-modal.component.html
```diff
@@ -0,0 +1,33 @@
...
+        <option [value]="option">{{ option }}</option>
+      }
+    </select>
+  </div>
+</div>
+<input #fileInput (change)="onImportClick($event)" id="file" type="file" accept=".csv" class="d-none" />
+<div class="modal-footer">
+  <div class="btn-group">
+    <button type="button" class="btn btn-primary" (click)="save()" id="save-button">Save</button>
+    <button type="button" class="btn btn-cancel" (click)="cancel()" id="cancel-button">Cancel</button>
```
burgerni10, line 31: Use internationalisation

burgerni10 (changes requested):
This seems to do the job but the logic could be eased by using the same endpoint to parse the data and check the correspondance. It will duplicate less code.

Also, you should use internationalisation on the frontend, let's talk about that

burgerni10 (approved):

PR #3568, opened 2024-09-10
chore: refactor scan mode model, controller, services and tests

burgerni10:


backend/src/service/scan-mode.service.ts
```diff
@@ -0,0 +1,73 @@
...
+    return this.scanModeRepository.findAll();
+  }
+
+  findById(id: string): ScanMode | null {
+    const scanMode = this.scanModeRepository.findById(id);
+    if (scanMode) {
+      return scanMode;
+    } else {
+      return null;
+    }
```
marouanehassanioptimistik, line 30: this is weird this should already be the return type of the repo method, you can just proxy to the repo call

backend/src/service/scan-mode.service.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  async delete(scanModeId: string): Promise<void> {
+    const scanMode = this.scanModeRepository.findById(scanModeId);
+    if (!scanMode) {
+      throw new Error(`Scan mode ${scanModeId} not found`);
+    }
+
+    this.scanModeRepository.delete(scanModeId);
+    this.southCacheRepository.deleteAllByScanMode(scanModeId);
```
marouanehassanioptimistik, line 61: do you have a foreign key? if yes then maybe the order matters not sure but this require a closer look

marouanehassanioptimistik (changes requested):
overall this is way better let's continue on this track some entities are more complex

backend/src/model/scan-mode.model.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export const toScanModeDTO = (scanMode: ScanMode): ScanModeDTO => {
+  return {
+    id: scanMode.id,
+    name: scanMode.name,
+    description: scanMode.description,
+    cron: scanMode.cron
+  };
+};
```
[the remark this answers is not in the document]
burgerni10: TODO: place it in the service

PR #3682, opened 2024-11-13
feat(oianalytics): manage South items import from OIAnalytics

burgerni10:


backend/src/model/oianalytics-command.model.ts
```diff
@@ -88,6 +88,16 @@ export interface OIBusDeleteNorthConnectorCommand extends BaseOIBusCommand {
...
 }
 
+export interface OIBusCreateOrUpdateSouthConnectorItemsFromCSVCommand extends BaseOIBusCommand {
+  type: 'create-or-update-south-items-from-csv';
+  southConnectorId: string;
+  commandContent: {
+    deleteItemsNotPresent: boolean;
+    csvContent: string;
+    delimiter: string;
+  };
```
marouanehassanioptimistik, line 98: it is not wrapped in a commandContent is it?

marouanehassanioptimistik (changes requested):

backend/src/model/oianalytics-command.model.ts
burgerni10: Not from OIAnalytics, but I wrap it in OIBus when received, in a similar fashion than OIBus update commands. Should I created dedicated fields instead (for both commands)?

marouanehassanioptimistik: not sure why you would transform the original signature, what does this bring?

marouanehassanioptimistik: nevermind it is oibus model forget what i said

marouanehassanioptimistik (approved):

Issue #3764, opened 2024-12-19
Improve test items result display


burgerni10:
Test items result are currently displayed in raw text or raw json.

The display of the results could be improved to have an array of values when retrieving time values

PR #3828, opened 2025-02-05
Improve test items result display

nagyszabi:
Close: #3764


burgerni10 (changes requested):
Small changes, otherwise ok !

burgerni10 (approved):

PR #3855, opened 2025-02-27
North rest api#3823

nagyszabi:


backend/src/north/north-rest/north-rest.ts
```diff
@@ -0,0 +1,259 @@
...
+      case 'time-values':
+        throw new Error('Can not manage time values');
+    }
+  }
+
+  /**
+   * Handle the file by sending it over to the specified endpoint
+   */
+  async handleFile(filePath: string): Promise<void> {
+    if (!(await filesExists(filePath))) {
```
burgerni10, line 46: To be coherent with the createReadStream, it should be path.resolve(filePath)

backend/src/north/north-rest/north-rest.ts
```diff
@@ -0,0 +1,259 @@
...
+      };
+    } finally {
+      if (!fileStream.closed) {
+        fileStream.close();
+      }
+    }
+
+    const ok = response.statusCode >= 200 && response.statusCode <= 299;
+    if (!ok) {
+      throw new Error(`HTTP request failed with status code ${response.statusCode} and message: ${await response.body.text()}`);
```
burgerni10, line 97: Let's generate an OIBus error like in OIAnalytics.

```typescript
throw new OIBusError(
    `Error ${response.status}: ${response.statusText}`,
    [401, 403, 404, 500, 502, 503, 504].includes(response.status)
  );
```

This will trigger a retry if it's a network failure. It doesn't retry an error otherwise (if badly formatted query for example - status 400). 
Maybe we should also include auth error to keep retrying the same data. Otherwise all the data will be added in error folder

burgerni10 (approved):

PR #3880, opened 2025-03-21
Replace fetch calls with undici request

nagyszabi:


backend/src/north/north-rest/north-rest.ts
```diff
@@ -61,26 +61,20 @@ export default class NorthREST extends NorthConnector<NorthRESTSettings> {
...
         query: queryParams,
         body: form,
-        dispatcher: proxyAgent,
-        signal: AbortSignal.timeout(this.connector.settings.timeout * 1000)
+        auth: authOptions,
+        proxy: proxyOptions,
+        timeout: this.connector.settings.timeout * 1000
       });
     } catch (error) {
       const message = this.getMessageFromError(error);
```
burgerni10, line 80: Is filestream properly closed, even if we throw an OIBus Error in the catch part?

backend/src/north/north-rest/north-rest.ts
```diff
@@ -61,26 +61,20 @@ export default class NorthREST extends NorthConnector<NorthRESTSettings> {
...
-      response = await request(endpoint, {
+      response = await HTTPRequest(endpoint, {
         method: 'POST',
         headers,
         query: queryParams,
         body: form,
-        dispatcher: proxyAgent,
-        signal: AbortSignal.timeout(this.connector.settings.timeout * 1000)
+        auth: authOptions,
+        proxy: proxyOptions,
```
burgerni10, line 76: Couldn't we set directly the headers, proxy and auth from calling the method instead of using a temp variable?

burgerni10: Same goes for testConnection

backend/src/service/http-request.utils.ts
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * URL is always returned, and is left unchanged when the auth type is not 'url'
+ */
+async function getAuthorization(options: ReqAuthOptions, url: string) {
+  const auth: { token?: string; url: string } = { url };
+
+  switch (options.type) {
+    case 'basic': {
+      const username = options.username;
+      const password = await encryptionService.decryptText(options.password ?? '');
```
burgerni10, line 95: With basic auth, password should be provided. Is this condition needed ? Instead, we can write options.password! since the encryption service now manage the empty string

backend/src/service/http-request.utils.ts
```diff
@@ -0,0 +1,141 @@
...
+      }
+
+      auth.token = token;
+      break;
+    }
+
+    case 'url': {
+      const username = options.username;
+      const password = await encryptionService.decryptText(options.password ?? '');
+
```
burgerni10, line 115: Same remark

backend/src/service/http-request.utils.ts
```diff
@@ -0,0 +1,141 @@
...
+      const urlObj = new URL(url);
+      urlObj.username = username;
+      urlObj.password = password;
```
burgerni10, line 118: I think these intermediate variables are not needed?

backend/src/north/north-oianalytics/north-oianalytics.ts
```diff
@@ -43,23 +39,26 @@ export default class NorthOIAnalytics extends NorthConnector<NorthOIAnalyticsSet
...
+    const authOptions = await this.getAuthorizationOptions();
+
+    const fetchOptions: ReqOptions = {
       method: 'GET',
-      headers: connectionSettings.headers,
-      timeout: this.connector.settings.timeout * 1000,
-      agent: connectionSettings.agent
+      auth: authOptions,
+      proxy: proxyOptions,
+      timeout: this.connector.settings.timeout * 1000
```
burgerni10, line 51: I don't think we need these intermediate variables, we can directly build and pass the options in the HTTPRequest call

backend/src/north/north-oianalytics/north-oianalytics.ts
```diff
@@ -77,25 +76,28 @@ export default class NorthOIAnalytics extends NorthConnector<NorthOIAnalyticsSet
...
-      },
-      timeout: this.connector.settings.timeout * 1000,
+      headers: { 'Content-Type': 'application/json' },
+      query,
       body: this.connector.settings.compress ? zlib.gzipSync(JSON.stringify(values)) : JSON.stringify(values),
-      agent: connectionSettings.agent
+      auth: authOptions,
+      proxy: proxyOptions,
+      timeout: this.connector.settings.timeout * 1000
     };
```
burgerni10, line 97: Same remark

backend/src/north/north-oianalytics/north-oianalytics.ts
```diff
@@ -105,8 +107,8 @@ export default class NorthOIAnalytics extends NorthConnector<NorthOIAnalyticsSet
 
     if (!response.ok) {
       throw {
-        message: `Error ${response.status}: ${response.statusText}`,
-        retry: [401, 403, 404, 500, 502, 503, 504].includes(response.status)
+        message: `Error ${response.statusCode}: ${await response.body.text()}`,
+        retry: [401, 403, 404, 500, 502, 503, 504].includes(response.statusCode)
       };
```
burgerni10, line 112: Let's build an OIBusError with the same logic than handleFile. Let's also use the same error code than North Rest for the retry.

backend/src/north/north-oianalytics/north-oianalytics.ts
```diff
@@ -140,22 +144,20 @@ export default class NorthOIAnalytics extends NorthConnector<NorthOIAnalyticsSet
...
     });
-    const formHeaders = body.getHeaders();
-    Object.keys(formHeaders).forEach(key => {
-      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-      // @ts-expect-error
-      connectionSettings.headers[key] = formHeaders[key];
-    });
+    const headers = body.getHeaders();
+    const proxyOptions = this.getProxyOptions();
+    const authOptions = await this.getAuthorizationOptions();
```
burgerni10, line 149: No need for these intermediate variables

backend/src/north/north-oianalytics/north-oianalytics.ts
```diff
@@ -172,104 +174,146 @@ export default class NorthOIAnalytics extends NorthConnector<NorthOIAnalyticsSet
 
     if (!response.ok) {
       throw new OIBusError(
-        `Error ${response.status}: ${response.statusText}`,
-        [400, 401, 403, 404, 500, 502, 503, 504].includes(response.status)
+        `Error ${response.statusCode}: ${await response.body.text()}`,
+        [400, 401, 403, 404, 500, 502, 503, 504].includes(response.statusCode)
```
burgerni10, line 178: Let's also use the same error code than North Rest for the retry.

burgerni10 (changes requested):

burgerni10 (approved):

Issue #3936, opened 2025-05-12
Deal better with import files that do not contain the right columns


matthieurolland:
When we try to load an old format OPC import file (3 columns) into the new OIBus version (many more columns), we get an error 500.

Wish : let the message explain that the file format does not match what is expected and may be quote the expected columns, ....

![Image](https://github.com/user-attachments/assets/1c7a55e9-a65a-4a98-a2b1-843c6cd986bf)

PR #4012, opened 2025-07-02
Deal better with import files that do not contain the right columns#3936

kbalinthunor:
UI before the changes (from issue #3936):
image

UI after the changes:
image


burgerni10 (changes requested):
Instead of displaying the error in a notification modal, could you display it in the import modal? And do not close the modal, to let the user reimport the file.

I think that this way the code can be mutualised between south and history query item imports

PR #4082, opened 2025-08-27
Rework of oibus ergonomics#4044

kbalinthunor:


frontend/src/app/north/north-subscriptions/north-subscriptions.component.html
```diff
@@ -19,19 +19,15 @@
...
+        @if (southConnectors.length !== (northConnector() ? subscriptions.length : subscriptions.length)) {
           <button class="btn btn-primary btn-sm oib-box-btn" (click)="addSubscription($event)" translate="north.subscriptions.add"></button>
         }
         <oib-help url="https://oibus.optimistik.com/docs/guide/north-connectors/common-settings#subscriptions" />
       </div>
     </div>
   </ng-template>
 
-  @if (northConnector() ? northConnector()!.subscriptions.length : subscriptions.length) {
+  @if (northConnector() ? subscriptions.length : subscriptions.length) {
```
burgerni10, line 30: Could be simplified

frontend/src/app/north/north-subscriptions/north-subscriptions.component.html
```diff
@@ -41,8 +37,8 @@
           </tr>
         </thead>
         <tbody>
-          @if (northConnector(); as northConnector) {
-            @for (subscription of northConnector.subscriptions; track subscription.id) {
+          @if (northConnector()) {
```
burgerni10, line 40: I think it can be removed

frontend/src/app/north/north-subscriptions/north-subscriptions.component.html
```diff
@@ -19,19 +19,15 @@
         }
       </div>
       <div>
-        @if (
-          northConnector()
-            ? southConnectors.length !== northConnector()!.subscriptions.length
-            : southConnectors.length !== subscriptions.length
-        ) {
+        @if (southConnectors.length !== (northConnector() ? subscriptions.length : subscriptions.length)) {
```
burgerni10, line 22: can be simplified (no need for northConnector anymore)

burgerni10 (changes requested):
I think there has been a misunderstanding in my issue: no need to create a new endpoint. "test settings" and "test connection" are the same thing. I wanted to say that when we click on the "Test settings" button, only the settings part of the form is validated. On the backend, we still use the same endpoint to test the connection

PR #4139, opened 2025-10-08
Drafr: Test custom transformer#4124

kbalinthunor:
UI after the changes:
image


burgerni10 (changes requested):
This looks good overall, but the user should be able to test a transformer in creation mode too.

I suggest to add a code template for new transformers, with the inputs arguments and returned outputs already set. The user will only have to manage the transform method content.

PR #4167, opened 2025-10-29
Export and import array on custom manifest#4144

### LINKED ISSUE #4144: Export and import array on custom manifest
For options with arrays, allow the user to export the items of the array (or an empty csv if no item).

The CSV will be generated the same way item exports works: it generate columns based on the manifest. Nested attributes are stringified.

Similarly, the user should be able to import elements into the array. Validation should be done dynamically on the frontend

kbalinthunor:
UI after the changes:
image
image
image
image
Closes #4144

### DIFF
--- a/frontend/src/app/history-query/history-query-items/history-query-items.component.spec.ts
+++ b/frontend/src/app/history-query/history-query-items/history-query-items.component.spec.ts
@@ -537,7 +537,6 @@ describe('HistoryQueryItemsComponent CSV Import Tests', () => {
 
     const mockModalRef = {
       componentInstance: {
-        expectedHeaders: [],
         prepare: jasmine.createSpy('prepare')
       },
       result: of({ file: new File([''], 'test.csv'), delimiter: ',' })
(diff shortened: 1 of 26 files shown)


backend/src/web-server/controllers/south-connector.controller.ts
```diff
@@ -521,4 +532,77 @@ export class SouthConnectorController extends Controller {
...
+    @Request() request: CustomExpressRequest
+  ): Promise<SouthArrayCsvImportResponse> {
+    const southService = request.services.southService as SouthService;
+
+    if (!file) {
+      throw new OIBusValidationError('Missing file "file"');
+    }
+    if (!delimiter) {
+      throw new OIBusValidationError('Missing delimiter');
+    }
```
burgerni10: Can be removed: validation is managed by the decorator

backend/src/service/south.service.ts
```diff
@@ -531,6 +533,166 @@ export default class SouthService {
...
+  ): Promise<{
+    items: Array<Record<string, unknown>>;
+    errors: Array<{ item: Record<string, string>; error: string }>;
+  }> {
+    if (!file) {
+      throw new OIBusValidationError('Missing file');
+    }
+    if (!delimiter) {
+      throw new OIBusValidationError('Missing delimiter');
+    }
```
burgerni10: no need to check that here

backend/src/service/south.service.ts
```diff
@@ -531,6 +533,166 @@ export default class SouthService {
...
+
+    const southConnector = this.findById(southId);
+    const existingItems = this.getArrayFieldItemsFromDatabase(southId, arrayKey);
```
burgerni10: let's check if we are in create mode (in this case, no need to retrieve from database)

burgerni10 (changes requested):
For this first review, I focused on the backend. I see many changes.

A global comment:
- let's use "element" instead of item, because item is used to refer to "item" data sources (item that we query in the data source)
- a lot can be simplified following the last refactor
- keep it modular: when this part will be approved, you will have to implement it for history queries, north, transformers too (everywhere we have manifest with possible array elements)

PR #4344, opened 2026-01-12
Item selection in transformer options #4332

### LINKED ISSUE #4332: Item selection in transformer options
On transformer options, when creating / editing the transformer attached to a north / history query, improve the item selection to:
- select all the items resulting from the search
- display the number of results in the search (just before the list for example, on top of "Random" on the example below - maybe you will have a better idea). Maybe the "select all result" can also be placed here? 
- limit the displayed result from the search when we have a very long list. Let's display the first ten items for example. That's why it's important to display the number of results.

Image

Add a button to remove all items from the option

Before the typeahead select, add a toggle to select all items (by default) or specific items. On specific items, it display the typeahead select.
When "select all items" is selected, the array of items attached to the options is empty by default.

kbalinthunor:
UI before the changes:
image
image

UI after the changes:
image
image
image

Closes #4332

### DIFF
--- a/frontend/src/app/history-query/history-query-transformers/edit-history-query-transformer-modal/edit-history-query-transformer-modal.component.html
+++ b/frontend/src/app/history-query/history-query-transformers/edit-history-query-transformer-modal/edit-history-query-transformer-modal.component.html
@@ -43,17 +43,19 @@
             </label>
           </div>
           @if (!selectAllItems) {
-            @if (searchInteracted && totalSearchResults > 0) {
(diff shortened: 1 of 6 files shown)


burgerni10 (changes requested):
Look good to me, but some things must be done:

- You forgot the "remove all" items when items are selected.
- The "select all items" and the number of result must be cleared once clicked
- When clicking in the search input, let's display the first items by default (even if the input is empty)

PR #4367, opened 2026-01-20
Arrange items by group in south connectors#4358

### LINKED ISSUE #4358: Arrange items by group in South connectors
On south items, add a group to arrange items together.

Grouped items can be queries together (in OPCUA or PI for example). They can also share some common settings, but these two things will be done later. First let's focus on the group feature (only for south items, not for history items).

On the item list, display the following things:
- Status
- Name
- Group
- Schedule
- Action buttons

Specific settings are not displayed anymore.

Image

When selecting items, Mass actions are grouped in a menu (next to the unselect all button). The menu title is "Mass actions (<numberOfSelectedItems> items)". The dropdown show the following action:
- Enable
- Disable
- Move to group
- Delete

The group can be selected from the edit modal, with a new select field just before the "Schedule" field.

Image

An item can be not attached to any group. In this case, the group selection display "None".

When selecting a group, the option list has a _special_ option "Create a new group" that will open a new modal with:
- the group name
- the default schedule for this group (can be override for each item)
- share tracked instant [only if the south connector has historian capabilities, like OPCUA, SQL, PI...]*
- default overlap (number of ms) [only if the south connector has historian capabilities, like OPCUA, SQL, PI...]*
*will be used later in the backend

The other options on the list when selecting a group have an edit button and a delete button. The delete button open a modal to ask for confirmation. On deletion, the items are moved back to the "None" group.

On the backend, it means a change in the SouthItem models and in the database with a two new tables:
- south_item_group (id, created_at, updated_at, name, south_id, scan_mode_id, share_tracked_instant)
   - the pair <name, south_id> must be unique
- group_items (group_id, item_id)
   - the pair must be unique

kbalinthunor:
UI after the changes:
South:

image
image
image
image

image
image
image

History:
image
image
image
image

Closes #4358


backend/shared/model/south-connector.model.ts
```diff
@@ -252,6 +257,103 @@ export interface ItemLightDTO extends BaseEntity {
...
+   * @example "periodic-5min"
+   */
+  scanModeId: string;
+
+  /**
+   * Whether items in this group share tracked instant.
+   *
+   * @example false
+   */
+  shareTrackedInstant: boolean;
```
burgerni10: Let's remove it for now.

It will be include in the item instead of the group (in a future PR). It was my mistake to ask you to put it here

backend/src/model/south-connector.model.ts
```diff
@@ -14,6 +14,16 @@ export interface SouthConnectorItemEntityLight extends BaseEntity {
   name: string;
 }
 
+export interface SouthItemGroupEntity extends BaseEntity {
+  name: string;
+  southId: string;
+  scanMode: ScanMode;
+  shareTrackedInstant: boolean;
```
burgerni10: Let's remove it for now.

It will be include in the item instead of the group (in a future PR). It was my mistake to ask you to put it here

PR #4389, opened 2026-02-03
South store last received value per item#4360

### LINKED ISSUE #4360: [South] Store last received value per item
For south connectors, use the cache.db to store the last value per item.

A similar logic is used for folder scanners, let's generalise that.

Let's use the cache.db database. Each connector will have a dedicated table (<connector-type>_<connector_id>) with the following fields:
- item_id
- query_time (instant)
- value (a JSON, content depends on the connector)
- tracked_instant (taken from cache history table)

This logic now makes the cache_history table useless: this table can be dropped.

The tracked instant value can be taken from the group tracked instant or individuel item tracked instant, depending on the logic selected with groups.

Tracked instant can be empty if the connector does not have history capabilities.

**Folder Scanner case**
Folder scanner already has a table <connector-type>_<connector_id>. Let's migrate the table by moving filename and mtime_ms column into a json array (filename, modifiedTime) by item. The associated item must be determined from the regex.
Then let's adapt the folder scanner to this logic.

**Frontend Display**
On the item list, add an action button to display these information on a modal.

kbalinthunor:
UI after the changes:
image
image
image

Closes #4360


burgerni10 (commented):
I still need to test it locally, but it looks good globally. The next step (in another PR) will be to rework the max instant management regarding this feature and the group feature.

Can you add the group attached to the item in the new modal?

PR #4665, opened 2026-05-29
Set default admin username, admin password and web server port from docker file and install script #4481

### LINKED ISSUE #4481: Set default admin username, admin password and web server port from docker file and install script
Let the user define the default username/password from docker env variables ADMIN_USERNAME, ADMIN_PASSWORD, and DEFAULT_PORT. Keep default value admin/pass and 2223.

I suggest these values to be passed as arguments to oibus process, and then passed down the user repository, so the default admin user is set with these values at startup.
Same for web server port passed down to the engine repository when initialising the engine settings.

When OIBus is installed from a script or from the windows installer, let the user choose these values and pass it down as arguments through the service.

kbalinthunor:
Closes #4481

### DIFF
--- a/backend/src/repository/config/user.repository.ts
+++ b/backend/src/repository/config/user.repository.ts
@@ -22,10 +22,10 @@ const DEFAULT_PASSWORD = 'pass';
 export default class UserRepository {
   constructor(
     private readonly database: Database,
-    private readonly defaultLogin = DEFAULT_USER.login,
-    private readonly defaultPassword = DEFAULT_PASSWORD
+    defaultLogin = DEFAULT_USER.login,
+    defaultPassword = DEFAULT_PASSWORD
   ) {
-    this.createDefault();
+    this.createDefault(defaultLogin, defaultPassword);
   }
 
   list(): Array<User> {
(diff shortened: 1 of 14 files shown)


backend/src/repository/config/user.repository.ts
```diff
@@ -20,7 +20,11 @@ const DEFAULT_USER: Omit<User, 'id' | 'createdBy' | 'updatedBy' | 'createdAt' |
 const DEFAULT_PASSWORD = 'pass';
 
 export default class UserRepository {
-  constructor(private readonly database: Database) {
+  constructor(
+    private readonly database: Database,
+    private readonly defaultLogin = DEFAULT_USER.login,
+    private readonly defaultPassword = DEFAULT_PASSWORD
```
burgerni10: It should not be kept in memory to avoid any password leak. Instead, let's just pass it down as arg. Same for login (you did well on engine)

burgerni10 (changes requested):
Since ServiceName is asked at the installation step, I think it can also be used to set the default OIBus name (in engine) at first startup.

Other minor changes, otherwise it seems ok, I'll test on Windows and docker before merging

burgerni10 (approved):

PR #4675, opened 2026-06-02
Influx db south connector#4478

### LINKED ISSUE #4478: Influx DB South Connector
Add a Influx DB south connector (database database).

InfluxDB has several package depending on the version. Let's implement all 3:
- v1: https://github.com/node-influx/node-influx
- v2: https://github.com/influxdata/influxdb-client-js/
- v3: https://github.com/InfluxCommunity/influxdb3-js

A setting in the connector should let the user select the version

kbalinthunor:
Closes #4478

### DIFF
--- a/backend/eslint.config.mjs
+++ b/backend/eslint.config.mjs
@@ -90,10 +90,12 @@ export default [
   {
     // Test files and shared mocks use async functions extensively to satisfy
     // Promise-returning interfaces, without ever needing an await expression.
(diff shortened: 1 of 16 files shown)


burgerni10 (changes requested):
I suggest a rework: let's use the same logic than South Rest API by returning raw array of records.

Results must then be converted with a json transformer on north connectors.

Keep the tracked instant logic from REST in the south. 

The serialization should now be gone.

FYI, in later versions, same logic will apply to all SQL connectors.

PR #4713, opened 2026-06-16
[Logger] - Improve logger feature #3689

### LINKED ISSUE #3689: [Logger] - Improve logger feature
Logger service should be a singleton, like the encryption service.

A rework of the pino logger should be done to properly use the appropriate scope (according to the log source) and properly update / shutdown the logger on settings changes.

kbalinthunor:
Closes #3689


burgerni10 (changes requested):
The purpose of this is also to directly use the singleton from the different components, instead of passing it in parameters everywhere. Let's change the rest of the code too.

Be careful to properly stop / update the right logger (when a connector name changed for example)

burgerni10 (approved):

PR #4733, opened 2026-06-22
Logger - Scope improvement #4714

### LINKED ISSUE #4714: Logger - Scope improvement
- Keep `internal` scope type, but move 'web-server' as a special scope id, used in pair with scope type 'internal'
- Create a new special scope id for 'engine', regarding logs from the engine and its various services
- Create two new fields `item_id` and `item_name` and use them when a log can be linked to items (a review on the south connectors logs can be done to do so)
- Display these fields in the log list
- Allow the user to search among the items (use the same suggest form field than south / north)

kbalinthunor:
Closes #4714


backend/src/service/oibus.service.ts
```diff
@@ -306,7 +306,7 @@ export default class OIBusService {
     await this.loggerService.stop();
     const registration = this.oIAnalyticsRegistrationService.getRegistrationSettings();
     await this.loggerService.start(settings, registration);
-    this.loggerEvent.emit('updated', this.loggerService.createChildLogger('web-server'));
+    this.loggerEvent.emit('updated', this.loggerService.createChildLogger('internal', 'web-server', 'Web Server'));
```
burgerni10: I think we can omit the scopeName when 'internal': instead we can use a translation key on the frontend on the different internal scope possibilities

backend/src/south/south-connector.ts
```diff
@@ -406,7 +406,8 @@ export default abstract class SouthConnector<T extends SouthSettings, I extends
           try {
             await this.directQueryHandler(groupedElements);
           } catch (error: unknown) {
-            this.logger.error(`Error when querying items with direct access: ${(error as Error).message}`);
+            const itemCtx = groupedElements.length === 1 ? { itemId: groupedElements[0].id, itemName: groupedElements[0].name } : {};
+            this.logger.error(itemCtx, `Error when querying items with direct access: ${(error as Error).message}`);
```
burgerni10: Seeing that, I think groups should be managed for connectors that are not "single run only" (modbus, opcua). I'm not sure of the design: should we have two other fields "groupId" and "groupName" or should we use the same fields ? Let me know what you can suggest

burgerni10 (changes requested):

backend/src/south/south-connector.ts
kbalinthunor: I think by using "groupId" and "groupName" will result in cleaner semantics, it will require some refactoring but the overall code quality will be much better

PR #4748, opened 2026-06-30
South offset for historian capable connectors #4741

### LINKED ISSUE #4741: South offset for historian capable connectors
An overlap already exists in South connectors.

Let's rework this feature:
- overlap should now be named "StartTime offset", still in ms, set to 0 by default
- a new field, called "EndTime offset", in ms, will be added, set to 0 by default
- StartTime offset is applid to the beginning of the history query interval. It can be applied as a positive or negative integer (frontend input must be adapted, and a migration must be done to manage the existing overlap as a negative offset)
- EndTime offset is applied to the end of the history query interval. If the end of the interval is the same or before the start, log a trace and do not execute the history query. EndTime will increase overtime (because it is computed as "now") and the history query will be trigger eventually
- This should be also included in group when history query is possible for south connector

kbalinthunor:
UI before the changes:
image
image

UI after the changes:
image
image

Closes #4741


backend/src/migration/entity-migrations/3/3.9/v3.9.2.ts
burgerni10, line 1: Add tests

burgerni10 (changes requested):

burgerni10 (approved):

PR #4761, opened 2026-07-06
OS specific fields #4757

### LINKED ISSUE #4757: OS specific fields
Add a new validator for manifest and in frontend dynamic-forms that display / enabled a field only on a specific platform (i.e. windows).

Dynamic forms should use the oibus service singleton to retrieve the info.

You can apply that to the newly reated fields for south folder scanner and north file writer (username, password, domain)

kbalinthunor:
Closes #4757

### DIFF
--- a/backend/shared/model/form.model.ts
+++ b/backend/shared/model/form.model.ts
@@ -26,6 +26,19 @@ export const OIBUS_ATTRIBUTE_TYPES = [
 export type OIBusAttributeType = (typeof OIBUS_ATTRIBUTE_TYPES)[number];
 
 /**
+ * List of platforms on which an attribute can be enabled.
+ * Values match the normalized platform exposed on OIBusInfo.platform.
+ */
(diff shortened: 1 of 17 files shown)


backend/shared/model/form.model.ts
```diff
@@ -193,6 +206,15 @@ export interface BaseOIBusDisplayableAttribute extends BaseOIBusAttribute {
...
   displayProperties: OIBusAttributeDisplayProperties;
+
+  /**
+   * Platforms on which this attribute is enabled.
+   * When absent, null or empty, the attribute is enabled on all platforms.
+   * When set, the attribute is only displayed/enabled and validated on the listed platforms.
+   *
+   * @example ["windows"]
+   */
+  enabledOnPlatforms?: Array<OIBusPlatform> | null;
```
burgerni10: Instead of a new field, I prefer to use a new validator

burgerni10 (changes requested):
I think it should be a validator instead of a new manifest field.

PR #4762, opened 2026-07-06
Use flags ignoreIpFilters and ignoreRemoteUpdate in the frontend to disable/hide fields #4758

### LINKED ISSUE #4758: Use flags ignoreIpFilters and ignoreRemoteUpdate in the frontend to disable/hide fields
When ignoreIpFilters is set, IP filters are disabled in the backend. To be coherent, the frontend should also not display them in this case, and display an information message instead.

Let's do the same thing with ignoreRemoteUpdate by setting to false (and disabling) the toggle in registration settings about remote update. Add an info message to explain why to the user

To use these flags on the frontend, I think it can be added to the OIBusInfo DTO and be retrieved from the service.

kbalinthunor:
Closes #4758

### DIFF
--- a/frontend/src/app/engine/ip-filter-list/ip-filter-list.component.ts
+++ b/frontend/src/app/engine/ip-filter-list/ip-filter-list.component.ts
@@ -1,11 +1,12 @@
 import { Component, inject, ChangeDetectionStrategy } from '@angular/core';
 
-import { firstValueFrom, switchMap } from 'rxjs';
+import { combineLatest, firstValueFrom, switchMap } from 'rxjs';
 import { Modal, ModalService } from '../../shared/modal.service';
 import { ConfirmationService } from '../../shared/confirmation.service';
 import { NotificationService } from '../../shared/notification.service';
 import { TranslateDirective, TranslatePipe } from '@ngx-translate/core';
 import { IpFilterService } from '../../services/ip-filter.service';
+import { EngineService } from '../../services/engine.service';
 import { IPFilterDTO } from '../../../../../backend/shared/model/ip-filter.model';
 import { EditIpFilterModalComponent } from './edit-ip-filter-modal/edit-ip-filter-modal.component';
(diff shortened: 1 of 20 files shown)


frontend/src/app/engine/ip-filter-list/ip-filter-list.component.ts
```diff
@@ -42,14 +43,19 @@ export class IpFilterListComponent {
...
   sortDirection: SortDirection = 'asc';
+  ignoreIpFilters = false;
 
   constructor() {
+    this.engineService.getInfo().subscribe(info => {
+      this.ignoreIpFilters = info.ignoreIpFilters;
+    });
     this.ipFilterService.list().subscribe(ipFilterList => {
       this.allIpFilters = ipFilterList;
       this.updateList(0);
```
burgerni10: Should be sync with combineLatest to load when both answer

frontend/src/app/engine/oia-registration/register-oibus-modal/register-oibus-modal.component.ts
```diff
@@ -123,6 +124,13 @@ export class RegisterOibusModalComponent {
       this.host = registration.host;
       this.form.controls.host.disable();
     }
+    this.oibusService.getInfo().subscribe(info => {
+      if (info.ignoreRemoteUpdate) {
+        this.ignoreRemoteUpdate.set(true);
+        this.form.controls.commandPermissions.controls.updateVersion.setValue(false);
+        this.form.controls.commandPermissions.controls.updateVersion.disable();
+      }
+    });
```
burgerni10: Since the flag is never updated, I think it should be passed as a boolean in the prepare method. No need to subscribe here like that, but it may be needed in the parent component

burgerni10 (changes requested):

burgerni10 (approved):

PR #4832, opened 2026-08-10
Improve History Query south progress monitoring for multi-item connectors #4829

### LINKED ISSUE #4829: Improve History Query south progress monitoring for multi-item connectors
The History Query detail page shows one progress bar for the south connector and one for the north connector. The south bar is driven purely by a time-interval ratio computed over the whole configured query time range (`intervalProgress` in `HistoryQueryMetrics.historyMetrics`).

For connectors that query items **one at a time** (all SQL-based connectors — MSSQL, MySQL, Oracle, PostgreSQL, SQLite, ODBC, OLE DB — plus folder-scanner, FTP, SFTP, REST, OIAnalytics), each item has its own independent time window and cache/tracked instant. Because the progress bar is computed from a single global interval list, it climbs toward 100% while item 1 runs, then **visibly jumps backward** as soon as the connector moves on to item 2 (which starts from its own, usually much earlier, tracked instant). With many items configured, the bar saws up and down repeatedly and gives no useful sense of overall completion.

There is also no indication anywhere in the UI of:
- which item is currently being queried,
- how many items have been completed vs. remain,
- or any per-item progress/status.

Connectors that batch all items into a single query call (OPC UA, PI, ADS, Modbus, MQTT) don't have the sawtooth problem since they share one timeline.

### Proposed improvement

Split south progress into two complementary indicators:

1. **Items progress bar** — "Item X / N" (count of items completed vs total), for connectors that query items individually. This directly reflects overall completion and never regresses.
2. **Current item progress bar** — the existing time-interval based bar, but scoped to the *current* item's own time range, so it fills 0→100% for the active item, then resets and starts filling again for the next item (rather than appearing to jump backward on a shared global scale).

For batched connectors (single call covering many items), the first bar isn't applicable, so only the current time-interval bar would show, as today — unless the connector can report finer-grained per-node/item completion internally, in which case it could adopt the same two-bar approach.

Suggested additions to make this concrete:
- Add `currentItemNumber` / `numberOfItems` (and optionally `currentItemName`) fields to `HistoryQueryMetrics.historyMetrics`, populated alongside the existing `history-query-interval` event emitted from `south-connector.ts`.
- Emit an item-boundary event (e.g. `history-query-item`) each time the single-items loop moves to a new item, carrying the item name/index.
- Frontend: render both bars in `history-metrics.component.html`, and label the current-item bar with the active item's name so users know what's being processed right now.

### Additional improvements to consider

- **Per-item status table**: expand the history query detail page with a small table listing each configured item and its status (pending / running / done), last value/timestamp retrieved, and records count — useful for diagnosing which specific item is slow or stuck, especially with dozens of items.
- **Rate/ETA feedback**: surface a values-or-files-per-second rate (derivable from existing `numberOfValuesRetrieved`/`numberOfFilesRetrieved` + elapsed time) and a rough ETA for the remaining items, since the sawtooth bar today gives no sense of "how much longer."
- **Persist last-completed item across restarts**: if a history query is paused/restarted mid-run, the UI could show it resuming from the correct item rather than restarting the visual progress from zero (the underlying per-item `trackedInstant` cache already supports this — it's a display gap, not a data gap).
- **Summary badge on the history query list view**: show "N/M items" or an aggregate percentage on the list page, not just in the detail page, so users can scan status across many history queries at once.

### Relevant code
- `backend/src/engine/history-query.ts` (`computeCurrentProgress`, `intervals`)
- `backend/src/south/south-connector.ts` (`historyQueryHandler`, `runHistoryQueryForLead`, `queryIntervals`)
- `backend/shared/model/engine.model.ts` (`HistoryQueryMetrics`)
- `backend/shared/model/south-connector.model.ts` (`SOUTH_SINGLE_ITEMS`)
- `frontend/src/app/history-query/history-query-detail/history-metrics/history-metrics.component.ts` / `.html`

kbalinthunor:
UI after the changes:

01-list
01-list-zoomed
02-detail

Closes #4829

### DIFF
--- a/backend/shared/model/engine.model.ts
+++ b/backend/shared/model/engine.model.ts
@@ -1552,7 +1552,92 @@ export interface HistoryQueryMetrics {
      * @example 2
      */
     numberOfIntervals: number;
+
+    /**
+     * The name of the item currently being queried. Only set for connectors that query items one
+     * at a time (SOUTH_SINGLE_ITEMS).
+     * @example "item1"
+     */
+    itemName?: string;
+
+    /**
+     * The 1-based index of the item currently being queried, among the run's enabled items. Only
+     * set for connectors that query items one at a time (SOUTH_SINGLE_ITEMS).
+     * @example 3
+     */
+    currentItemNumber?: number;
+
+    /**
+     * The total number of enabled items in the run. Only set for connectors that query items one
+     * at a time (SOUTH_SINGLE_ITEMS).
+     * @example 10
+     */
+    numberOfItems?: number;
+
+    /**
+     * The progress of the current item's own interval list as a fraction [0, 1], scoped to that
+     * item only (not ratcheted, unlike `intervalProgress`). Only set for connectors that query
+     * items one at a time (SOUTH_SINGLE_ITEMS).
+     * @example 0.25
+     */
+    itemIntervalProgress?: number;
+
+    /**
+     * The 1-based index of the current lead's own interval, within its own interval list (raw, not
+     * ratcheted) — same source data as `itemIntervalProgress`, surfaced raw for display. Only set
+     * for connectors that query items one at a time (SOUTH_SINGLE_ITEMS).
+     * @example 12
+     */
+    itemIntervalNumber?: number;
+
+    /**
+     * The total number of intervals in the current lead's own interval list (raw, not ratcheted).
+     * Only set for connectors that query items one at a time (SOUTH_SINGLE_ITEMS).
+     * @example 34
+     */
+    itemNumberOfIntervals?: number;
+
+    /**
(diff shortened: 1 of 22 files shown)


burgerni10:
I'm not sure to understand why there is the three dots in the history list ?
Also, on your screenshots, I think you can simplify by displaying and ordering the following metrics:
- Items progress. Display the progress bar at the end of the line. Something like: Current item: <item name> (2 / 4) <progress bar>
- Current item interval progress. Display the content of the interval progress here, followed by the progress bar
- I think "data retrieved" and "interval progress" are not useful anymore.
- Your second screenshot seem inconsistent between the interval progress and the item progress

You can also rebase from 3.10

burgerni10 (approved):

PR #4861, opened 2026-08-24
Add an audit trail for configuration changes (create/update/delete with before/after snapshots) #4856

### LINKED ISSUE #4856: Add an audit trail for configuration changes (create/update/delete with before/after snapshots)
## Problem

OIBus entities (south/north connectors, scan modes, IP filters, certificates, users, history queries, transformers, OIAnalytics registration, etc.) already track `created_by`/`created_at`/`updated_by`/`updated_at` on each row, but only the *latest* state and the *identity of the last editor* are known. There's no way to answer "what did this connector's config look like before it broke?" or "who deleted this scan mode, and what was in it?" — the previous state is simply overwritten or gone.

### 1. New `audit_logs` table

A dedicated table, independent of the entity tables, recording every create/update/delete across OIBus:

| column | type | notes |
|---|---|---|
| `id` | string (PK) | |
| `entity_type` | string | e.g. `south_connector`, `north_connector`, `scan_mode`, `ip_filter`, `certificate`, `user`, `history_query`, `transformer`, `engine`, `oianalytics_registration` |
| `entity_id` | string | id of the affected row |
| `action` | string | `CREATE` \| `UPDATE` \| `DELETE` |
| `previous_state` | json (nullable) | full-row snapshot before the change, **excluding the audit bookkeeping fields themselves** (`created_at`, `created_by`, `updated_at`, `updated_by`); null for `CREATE` |
| `new_state` | json (nullable) | same, for the state after the change; null for `DELETE` |
| `user_id` | string | the authenticated user who made the change (mirrors the existing `updated_by`/`created_by` convention) |
| `created_at` | datetime | when the change was recorded |

Added via a new entity migration under `backend/src/migration/entity-migrations/3/3.10` (next minor version folder).

**Nested child records** (e.g. south items within a south connector, history items within a history query, north transformer links) are logged as **their own separate `audit_logs` rows** (own `entity_type`/`entity_id`), not folded into the parent entity's snapshot.

**Scope of tracked entities:** all entities that currently carry `created_by`/`updated_by` — south/north connectors (+ items/groups), history queries (+ items), scan modes, IP filters, certificates, users, transformers, engine settings, and OIAnalytics registration.

Only changes made by an authenticated user through the UI/API or through the OIAnalytics command feature are recorded, not the migration or system changes.

### 2. Backend plumbing

- `AuditRepository` in `backend/src/repository/config/`, following the existing raw-SQL repository style.
- **Centralized hook:** introduce a thin shared wrapper/service that repositories call into for create/update/delete, rather than each of the ~10 repositories duplicating snapshot-and-diff logic inline. Concretely: an `AuditService.record(entityType, entityId, action, previousState, newState, userId)` helper, invoked from each repository's existing create/update/delete methods, with the audit-field stripping (`created_at`/`created_by`/`updated_at`/`updated_by`) handled once inside that helper so no repository has to remember to do it. This is still a refactor touching every tracked repository, but keeps the stripping/serialization logic in one place instead of ~10 copies.
- `AuditController` (`backend/src/web-server/controllers/`) exposing `@Route('/api/audit')` with:
  - `GET /api/audit` — paginated/filterable list (by `entity_type`, `entity_id`, `action`, date range) for the global Audit page.
  - `GET /api/audit/{entityType}/{entityId}` — history for a single entity, for the per-entity audit button.
- Retention: a configurable retention setting (similar to existing log/metrics retention) plus a scheduled cleanup job that prunes `audit_logs` rows older than the configured window. The retention field can be added to the engine, and displayed /set in the logger section (that can be renamed Logging and audit).

### 3. Frontend

- **Per-entity audit button**: add a third icon button (e.g. `fa-history`) next to the existing edit/delete buttons in each entity list's `.pull-right` action-buttons group (e.g. `south-list.component.html`, `north-list.component.html`, `scan-mode-list.component.html`, `ip-filter-list/`, `certificate-list/`, `transformer-list/`, user list, history-query list). Opens a modal showing that entity's audit history (before/after diff per change), including its child records' own audit rows where relevant (e.g. a south connector's item changes).
- **Global "Audit" page**: new route (e.g. `/audit`), listing all audit events across all entity types, filterable/sortable, reusing the same before/after diff view.
- **User menu entry**: add an "Audit" link in `navbar.component.html`'s account dropdown, alongside the existing Settings/Logout items (both the desktop and mobile dropdown blocks).
- Reuse or extend the existing `oib-audit-info` shared component (`frontend/src/app/shared/audit-info/`) where sensible, since it already surfaces created/updated metadata in list tables today.

## Acceptance criteria

- [ ] `audit_logs` table created via migration, with JSON snapshot columns.
- [ ] Every create/update/delete of a tracked entity (and its tracked child records, as separate rows) writes exactly one audit row with correct `previous_state`/`new_state`, excluding audit bookkeeping fields from the snapshots.
- [ ] Snapshot/diff logic centralized in a shared `AuditService` called from each repository, not duplicated per repository.
- [ ] `GET /api/audit` and `GET /api/audit/{entityType}/{entityId}` implemented and documented via tsoa/OpenAPI.
- [ ] Configurable retention + scheduled pruning job.
- [ ] Audit icon button added to each tracked entity's list view, opening that entity's history.
- [ ] Global Audit page reachable from the user menu, listing/filtering all events.
- [ ] Unit tests for the repository/service (snapshot correctness, field exclusion, retention pruning) and frontend component tests for the new views.

kbalinthunor:
UI changes:
01-navbar-account-dropdown
02-ip-filter-list-with-audit-button
03-global-audit-page-populated
04-audit-history-modal
05-audit-diff-expanded
06-logging-and-audit-modal

Closes #4856

### DIFF
--- a/backend/shared/model/audit.model.ts
+++ b/backend/shared/model/audit.model.ts
@@ -1,0 +1,67 @@
+export type AuditAction = 'CREATE' | 'UPDATE' | 'DELETE';
+
+export type AuditEntityType =
+  | 'south_connector'
+  | 'south_item'
+  | 'south_item_group'
+  | 'north_connector'
+  | 'north_transformer'
+  | 'history_query'
+  | 'history_query_item'
+  | 'history_query_transformer'
+  | 'scan_mode'
+  | 'ip_filter'
+  | 'certificate'
+  | 'user'
+  | 'transformer'
+  | 'engine'
+  | 'oianalytics_registration';
+
+/**
+ * Data Transfer Object for an audit log entry.
+ * Represents a single create/update/delete event recorded against an audited entity.
+ */
+export interface AuditLogDTO {
+  /**
+   * The unique identifier of the audit log entry.
+   */
+  id: string;
+
+  /**
+   * The type of entity this audit log entry relates to.
+   * @example "south_connector"
+   */
+  entityType: AuditEntityType;
+
+  /**
+   * The identifier of the entity this audit log entry relates to.
+   */
+  entityId: string;
+
+  /**
+   * The kind of change performed on the entity.
+   * @example "UPDATE"
+   */
+  action: AuditAction;
+
+  /**
+   * A JSON snapshot of the entity before the change. `null` for `CREATE`.
+   */
+  previousState: Record<string, unknown> | null;
+
+  /**
+   * A JSON snapshot of the entity after the change. `null` for `DELETE`.
+   */
+  newState: Record<string, unknown> | null;
+
+  /**
+   * The identifier of the user who performed the change.
+   */
+  userId: string;
+
+  /**
+   * ISO timestamp of when the change was recorded.
+   * @example "2026-01-01T00:00:00.000Z"
+   */
+  createdAt: string;
+}
--- a/backend/shared/model/audit.model.ts
+++ b/backend/shared/model/audit.model.ts
@@ -1,21 +1,38 @@
-export type AuditAction = 'CREATE' | 'UPDATE' | 'DELETE';
+/**
+ * List of possible audit actions.
+ */
+export const AUDIT_ACTIONS = ['CREATE', 'UPDATE', 'DELETE'] as const;
+/**
+ * Type representing an audit action.
+ * @example 'UPDATE'
+ */
+export type AuditAction = (typeof AUDIT_ACTIONS)[number];
 
-export type AuditEntityType =
-  | 'south_connector'
-  | 'south_item'
-  | 'south_item_group'
-  | 'north_connector'
-  | 'north_transformer'
-  | 'history_query'
-  | 'history_query_item'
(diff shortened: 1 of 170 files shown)


burgerni10 (changes requested):
Backend is ok, but I see some improvements on the UI:
- place audit icon / button on the far right of the icon list
- audit history should display three modes: 
   - display (your view in an array)
   - json diff
   - json side by side

Diff mode:
Image

Side by side mode is both json, side by side.

Also, I think the audit history modal, and the logging and audit modal, can be increased in size.

kbalinthunor:
Updated UI components:
01-south-list-icon-order
02-audit-modal-xl-size
03-audit-table-mode
04-audit-json-diff-mode
05-audit-json-side-by-side-mode
06-logger-modal-lg-size

PR #4877, opened 2026-08-26
New MongoDB South connector #4867

### LINKED ISSUE #4867: New MongoDB South connector
## Summary

Add a MongoDB south connector, so OIBus can pull time-series/event data out of a MongoDB collection on a schedule, the same way it already does for PostgreSQL, MySQL, MSSQL, Oracle, and SQLite.

## Proposed solution

Follow the existing SQL south connector pattern (`backend/src/south/south-postgresql/`, `south-mysql/`, `south-mssql/`) closely — MongoDB fits the same shape (`category: 'database'`, `modes: { history: true }` only, no subscription/lastPoint/lastFile), just with Mongo's own query language instead of SQL.

### Connector-level settings (manifest)

Follow the `south-odbc` precedent (`backend/src/south/south-odbc/manifest.ts`), not the `south-postgresql` host/port precedent — MongoDB's real-world topology (replica sets, `mongodb+srv://` Atlas seed lists, sharded clusters, arbitrary URI options) doesn't decompose cleanly into a single host/port pair the way a standalone RDBMS endpoint does, and forcing it into discrete fields would mean rebuilding, in OIBus's UI, a set of connection shapes MongoDB's own URI syntax already covers.

- `connectionString`: everything *except* credentials — e.g. `mongodb://host1:27017,host2:27017,host3:27017/database?replicaSet=rs0&authSource=admin` or `mongodb+srv://cluster.example.net/database?...`. Default value shown as a template, same idea as ODBC's placeholder. `displayInViewMode: true`, since it never contains a secret.
- `username`: plain string field (not secret — matches every other south's convention of a plain `username` alongside a secret `password`).
- `password`: `type: 'secret'`, `displayInViewMode: false`.
- `connectionTimeout`.

### Item-level settings

Each item targets one collection with either a filter query or an aggregation pipeline:
- `collection` (string)
- `queryType`: `'find'` or `'aggregation'` — `find` covers the common case (a JSON filter object), `aggregation` covers pipelines needing `$match`/`$project`/`$group` for anything beyond a flat filter.
- `query`: a JSON string (the filter object, or the pipeline array for aggregation mode) — same idea as the existing `query` field in `south-postgresql`'s item settings, just JSON instead of SQL text.
- Incremental time-range querying reuses the existing `@StartTime`/`@EndTime` placeholder convention every SQL south already uses (see `south-postgresql.ts`'s `queryData()`), substituted into the filter/pipeline JSON as ISO date strings before `JSON.parse`, e.g. `{ "timestamp": { "$gt": "@StartTime", "$lte": "@EndTime" } }` — converted to a BSON `Date` after parsing, the same way the SQL connectors convert their placeholders into dialect-appropriate parameters.
- A `sort` field (defaulting to the same timestamp field used in the query) so results arrive in a well-defined order for the incremental cursor, mirroring how the SQL connectors rely on `@StartTime`/`@EndTime` plus row order for pagination.

### Driver

The official `mongodb` npm driver. Unlike `oracledb` or `better-sqlite3`, it has no native binding to bundle — lower packaging risk for `@yao-pkg/pkg` (no new `pkg.assets`/`pkg.scripts` entry expected, though this should be confirmed while implementing).

### Testing

- Unit tests mocking the driver, following the existing `south-postgresql.spec.ts` structure.
- Add a `mongo` service to `docker-compose.yml` under the existing `database` profile (alongside `postgres`/`influxdb`), so it's available via `npm run docker:dev` for manual testing — and so it can be picked up by the CI integration-testing work (both the connector-level and bundle-level proposals already drafted) without a separate infrastructure change later.

## Out of scope (for v1)

- Change streams / real-time subscription mode — v1 is poll-based history query only, matching every other database south.
- X.509 certificate authentication — username/password only for v1.
- Sharded-cluster-specific configuration beyond a standard connection string.

## Acceptance criteria

- [ ] New `south-mongodb` connector following the existing south connector structure (`manifest.ts`, `south-mongodb.ts`, spec file).
- [ ] Connects via a credential-free `connectionString` field (supporting standard and `mongodb+srv://` URIs, replica sets, and arbitrary URI options) plus separate `username`/`password` fields, with `password` passed via the driver's native `auth` option rather than embedded in the string.
- [ ] Item settings support both a plain filter query and an aggregation pipeline against a chosen collection.
- [ ] `@StartTime`/`@EndTime` placeholders work the same way they do in the existing SQL south connectors, for incremental history queries.
- [ ] `docker-compose.yml` gains a `mongo` service under the `database` profile for local/CI testing.
- [ ] Documentation page added alongside the other database south connectors.

kbalinthunor:
UI after the changes:

01-add-south-type-list
02-mongodb-connector-settings
03-mongodb-item-settings
04-mongodb-item-saved-list

Closes #4867


backend/src/south/south-mongodb/south-mongodb.ts
```diff
@@ -0,0 +1,280 @@
...
+    for (const doc of docs) {
+      const rawValue = doc[fieldName];
+      if (rawValue === null || rawValue === undefined) continue;
+      const instant = String(rawValue) as Instant;
+      if (!updatedStartTime || instant > updatedStartTime) {
+        updatedStartTime = instant;
+      }
+    }
+    return updatedStartTime;
+  }
```
burgerni10: Wouldn't it be better to a the same jsonpath logic than the one used in transformers to retrieve the max instant from that path ?

burgerni10 (changes requested):
I'm not sure the trackMaxInstant s the right approach

backend/src/south/south-mongodb/south-mongodb.ts
kbalinthunor: I agree on this, implemented the suggested solution

burgerni10 (approved):

